PR #40, opened 2020-06-06
Support for non-blocking and non-JDBC persistence

### LINKED ISSUE #15: [feature request] Non-blocking support (R2DBC)

badgerwithagun:
Fully extends the SPI to allow for any theoretical data store, SQL or otherwise, blocking or non-blocking, and adds functioning R2DBC support.

Greatly expands test coverage via introduction of a unified test suite. Any combination of database, `Persistor` and `TransactionManager` implementation can be run through the same set of core acceptance tests.

Fixes #15

And lots of other things....


RomainWilbert:
HI @badgerwithagun !

Can we expect a release with R2DBC support soon ?

badgerwithagun:
I keep getting close on this and then chickening out.

I need to bring this branch back up-to-date. It's a complete refactor with almost every aspect changed to support asynchrony, so merging from master is a lot of work.

The final bit of testing I need to do is to test it with "real world" frameworks like the latest R2DBC-supporting spring boot versions.

Now I know someone's interested, I'll spend some time on it soon.

lperdigaomirriad:
It would be great to see this coming to the library :)

RomainWilbert:
We have a use case where project would like to use Redis as main database. I think the refactor provided by this PR would help to support non-SQL databases.

@badgerwithagun Would you say the effort could be huge to support Redis ?

badgerwithagun:
@RomainWilbert I don't know enough about Redis to say.  I know it has transactions, but whether they could map to the same concepts of "connections" and "transactions" and "commands" as SQL is probably one you can better answer!

You have reminded me that this is an itch I really need to scratch. Now R2DBC is a first-class feature of Spring, I should be able to do some much more meaningful tests on all this and get closer to merging it with the mainline.

badgerwithagun:
Taken the best ideas from this and implemented them elsewhere.

With virtual threads now a reality, there is no reason to implement R2DBC

PR #43, opened 2020-06-09
Use Spring's utilities to retrieve transactions

### LINKED ISSUE #34: transactionoutbox-spring depends on hibernate-core
I am using Spring, but not Hibernate, and the code in SpringTransactionManager won't work for me. There is no `SessionFactory` from which to fetch the active `Connection`.

I _think_ the answer is just to inject the DataSource, and use connections from that. You should be able to require that this DataSource has been either
a) wrapped in [TransactionAwareDataSourceProxy](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/jdbc/datasource/TransactionAwareDataSourceProxy.html) in a non-JTA context
b) is a JTA connection that will handle the transactionality for you.

If this make sense, I might be able to create a PR.

tsg21:
Fixes #34 by using Spring utilities to get the JDBC Connection.

### DIFF
--- a/transactionoutbox-spring/pom.xml
+++ b/transactionoutbox-spring/pom.xml
@@ -34,6 +34,11 @@
       <version>${spring.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-jdbc</artifactId>
+      <version>${spring.version}</version>
+    </dependency>
+    <dependency>
       <groupId>javax.inject</groupId>
       <artifactId>javax.inject</artifactId>
       <version>1</version>
@@ -42,11 +47,6 @@
       <groupId>org.springframework</groupId>
       <artifactId>spring-context</artifactId>
       <version>${spring.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>org.hibernate</groupId>
-      <artifactId>hibernate-core</artifactId>
-      <version>5.4.17.Final</version>
     </dependency>
 
     <!-- Compile -->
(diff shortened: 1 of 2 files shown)


transactionoutbox-spring/src/main/java/com/gruelbox/transactionoutbox/SpringTransactionManager.java
```diff
@@ -43,8 +47,11 @@
     @Override
     public Connection connection() {
       try {
-        SessionImpl delegate = (SessionImpl) entityManager.getDelegate();
-        return delegate.connection();
+        if (!TransactionSynchronizationManager.isActualTransactionActive()) {
```
badgerwithagun, line 50: A better place for this check would be in `requireTransactionReturns`; then `NoTransactionActive` exception can be thrown correctly in line with the contract.

This `try..catch` block should be removed, I think.

badgerwithagun:
One of the things I'm in the process of doing on #33 is creating a standardised test suite that tests all SPI features against the core contracts. AT the moment there's nothing doing edge or stress testing against this transaction manager.

That said, it's `spring-tx`. It Just Works, in theory (notwithstanding the Spring AOP vs AspectJ kerfuffle, needing `public` methods and all that nonsense).  I'm not sure now much testing it really needs ;)

badgerwithagun:
I think the build fell over due to some temporary Maven issue. Kicked it off again.

transactionoutbox-spring/pom.xml
```diff
@@ -43,11 +48,6 @@
       <artifactId>spring-context</artifactId>
```
badgerwithagun, line 48: Still need this?

tsg21:
> That said, it's `spring-tx`. It Just Works, in theory (notwithstanding the Spring AOP vs AspectJ kerfuffle, needing `public` methods and all that nonsense). I'm not sure now much testing it really needs ;)

Yes, I spent a while staring at your existing tests and decided they had enough coverage for this change.

transactionoutbox-spring/pom.xml
tsg21: Checked, yes, it contains the `@Service` annotation.

badgerwithagun:
Looks great! Thanks @tsg21! I'll put out a release.

PR #53, opened 2020-06-20
Scope fixes for the Spring extension

### LINKED ISSUE #44: transactionoutbox-spring depends on specific Spring version
At the moment, it depends on Spring 5.2.6, but there's no guarantee that this will be the version that is run by the consumer.

Shouldn't the spring dependency (and others like it) be set to `<scope>provided/<scope>`?

Docs: [Dependency Scopes](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scopel)

That way the testing will be done against a specific version, but it's pulled in by consumers of the lib, and it's not transitive.

This then behaves a little like an npm "peer dependency".

Or is there a better way to achieve this that I am not aware of?

tsg21:
Fixes #44

### DIFF
--- a/transactionoutbox-spring/pom.xml
+++ b/transactionoutbox-spring/pom.xml
@@ -34,6 +34,11 @@
       <version>${spring.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-jdbc</artifactId>
+      <version>${spring.version}</version>
+    </dependency>
+    <dependency>
       <groupId>javax.inject</groupId>
       <artifactId>javax.inject</artifactId>
       <version>1</version>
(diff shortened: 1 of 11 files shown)


transactionoutbox-spring/pom.xml
```diff
@@ -59,18 +62,22 @@
     <dependency>
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
+      <scope>test</scope>
```
tsg21, line 65: Were these dependencies deliberately not "test"?

badgerwithagun: They're set to `test` in the parent `dependencyManagement`.

https://github.com/gruelbox/transaction-outbox/blob/master/pom.xml#L129

badgerwithagun:
@tsg21, It'd be really useful if you could spare a moment to give `2.1.13-SNAPSHOT` in the Sonatype snapshots repository (OSSRH) a try.  It's a build of #40 which, although primarily a refactor to a non-blocking core, makes a number of (what I think are) improvements to the Spring side of things which would affect you, and I lack real-world apps to test against.

badgerwithagun:
*bump* @tsg21

PR #70, opened 2020-09-15
:') no shared alter syntax - adding dialect specific option to migrat…

victokoh:
…ions.

### DIFF
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
@@ -4,6 +4,7 @@ import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
```diff
@@ -39,27 +43,37 @@
           new Migration(
               4,
               "Add flush index",
-              "CREATE INDEX IX_TXNO_OUTBOX_1 ON TXNO_OUTBOX (processed, blacklisted, nextAttemptTime)"));
+              "CREATE INDEX IX_TXNO_OUTBOX_1 ON TXNO_OUTBOX (processed, blacklisted, nextAttemptTime)"),
```
badgerwithagun, line 46: I would have thought `ALTER` is the default, no?

victokoh: not too sure what you mean? the default specified here is for the mysql dialects, since it looked like postgres9 and h2 had slightly different syntax for alter anyway.

badgerwithagun: Ah yes, does look like modify is more common. Learned a thing :)

badgerwithagun (approved):

PR #116, opened 2021-02-28
+1 add last attempt time to outbox entries. null on initial insert, c…

victokoh:
…urrent time @ clock provider for subsequent updates.

### DIFF
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
@@ -64,7 +64,7 @@ class DefaultMigrationManager {
           new Migration(
               7,
               "Add lastAttemptTime column to outbox",
-              "ALTER TABLE TXNO_OUTBOX ADD COLUMN lastAttemptTime TIMESTAMP(6) AFTER invocation",
+              "ALTER TABLE TXNO_OUTBOX ADD COLUMN lastAttemptTime TIMESTAMP(6) NULL AFTER invocation",
               Map.of(
                   Dialect.POSTGRESQL_9,
                   "ALTER TABLE TXNO_OUTBOX ADD COLUMN lastAttemptTime TIMESTAMP(6)")));
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
@@ -7,7 +7,6 @@ import java.sql.Statement;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import javax.validation.constraints.NotNull;
 import lombok.AllArgsConstructor;
 import lombok.SneakyThrows;
@@ -20,7 +19,9 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 class DefaultMigrationManager {
 
-  /** Migrations can be dialect specific **/
+  /**
+   * Migrations can be dialect specific *
+   */
   private static final List<Migration> MIGRATIONS =
       List.of(
           new Migration(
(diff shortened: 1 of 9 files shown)


transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
```diff
@@ -60,7 +60,14 @@
                   Dialect.POSTGRESQL_9,
                   "ALTER TABLE TXNO_OUTBOX RENAME COLUMN blacklisted TO blocked",
                   Dialect.H2,
-                  "ALTER TABLE TXNO_OUTBOX RENAME COLUMN blacklisted TO blocked")));
+                  "ALTER TABLE TXNO_OUTBOX RENAME COLUMN blacklisted TO blocked")),
+          new Migration(
+              7,
+              "Add lastAttemptTime column to outbox",
+              "ALTER TABLE TXNO_OUTBOX ADD COLUMN lastAttemptTime TIMESTAMP(6) AFTER invocation",
+              Map.of(
```
victokoh, line 68: https://wiki.postgresql.org/wiki/Alter_column_position

badgerwithagun:
Looking great! Just needs the a test or two.

victokoh:
> Looking great! Just needs the a test or two.

added.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
```diff
@@ -60,7 +60,14 @@
                   Dialect.POSTGRESQL_9,
                   "ALTER TABLE TXNO_OUTBOX RENAME COLUMN blacklisted TO blocked",
                   Dialect.H2,
-                  "ALTER TABLE TXNO_OUTBOX RENAME COLUMN blacklisted TO blocked")));
+                  "ALTER TABLE TXNO_OUTBOX RENAME COLUMN blacklisted TO blocked")),
+          new Migration(
+              7,
+              "Add lastAttemptTime column to outbox",
+              "ALTER TABLE TXNO_OUTBOX ADD COLUMN lastAttemptTime TIMESTAMP(6) NULL AFTER invocation",
```
victokoh, line 67: got caught out by: https://dev.mysql.com/doc/refman/5.7/en/timestamp-initialization.html 
*explicit null declaration required on new column when dealing with mysql5 otherwise auto-update is in effect.

transactionoutbox-core/src/test/java/com/gruelbox/transactionoutbox/TestDefaultPersistorMySql8.java
```diff
@@ -34,6 +34,6 @@ protected TransactionManager txManager() {
 
   @Override
   protected Dialect dialect() {
-    return Dialect.MY_SQL_5;
+    return Dialect.MY_SQL_8;
```
victokoh, line 37: seemed like a bug? updated.

badgerwithagun:
> > Looking great! Just needs the a test or two.
> 
> added.

I think this is the key one we need a test in: https://github.com/gruelbox/transaction-outbox/blob/master/transactionoutbox-core/src/test/java/com/gruelbox/transactionoutbox/acceptance/AbstractAcceptanceTest.java

victokoh:
> > > Looking great! Just needs the a test or two.
> > 
> > 
> > added.
> 
> I think this is the key one we need a test in: https://github.com/gruelbox/transaction-outbox/blob/master/transactionoutbox-core/src/test/java/com/gruelbox/transactionoutbox/acceptance/AbstractAcceptanceTest.java

Added! Let me know if there's anything else you'd like me to modify.

On an unrelated note, I noticed something strange on the version value whilst writing this test. Couldn't quite figure it out at this time, but would be interested to know if you think it's indeed some kind of separate bug (*see version on entry 2):

image

transactionoutbox-core/src/test/java/com/gruelbox/transactionoutbox/acceptance/AbstractAcceptanceTest.java
```diff
@@ -406,6 +402,48 @@ final void retryBehaviour() throws Exception {
         });
   }
 
+    @Test
```
victokoh, line 405: think my formatting's off from yours - 😬 do you have a formatting config file that can be checked in?

badgerwithagun: Just run `mvn coveo:fmt`.

badgerwithagun:
> On an unrelated note, I noticed something strange on the version value

Hm, doesn't make sense on paper. Something to look at separately? LGTM for now

badgerwithagun (approved):

Issue #134, opened 2021-04-16
Connections not closed with SpringTransactionManager


GrannyOlli:
Hi,

thanks for your great library!

I tried it out using the spring-integration. I noticed that db-connections won't be given back to the connection pool if a callback throws an exception. This leads to the situation, that after some failed callbacks no connections are available any more.

The problem seems to be that the default methods of com.gruelbox.transactionoutbox.TransactionManager are not wrapped by Springs TransactionInterceptor, because they are not directly annotated with @Transactional.

Example: Method inTransactionThrows called in TransactionOutboxImpl.updateAttemptCount is not wrapped by Springs transaction-aspect and thus connections are not given back to the connection pool after the method ends. inTransactionThrows calls inTransactionReturnsThrows, which is not intercepted by the aop-proxy when called from within the object, only when called directly.

Copying all default methods to SpringTransactionManager and annotating them with @Transactional solves the problem in my test setup.

Example:

@Beta
@Service
public class SpringTransactionManager implements ThreadLocalContextTransactionManager {
        ...
	@Override
	@Transactional(propagation = Propagation.REQUIRES_NEW)
	public void inTransaction(Runnable runnable) {
		uncheck(() -> inTransactionReturnsThrows(ThrowingTransactionalSupplier.fromRunnable(runnable)));
	}

	@Override
	@Transactional(propagation = Propagation.REQUIRES_NEW)
	public void inTransaction(TransactionalWork work) {
		uncheck(() -> inTransactionReturnsThrows(ThrowingTransactionalSupplier.fromWork(work)));
	}
....

Best regards
Oliver

badgerwithagun:
Thanks @GrannyOlli ! Would you consider opening a pull request with this change?

We don't use Spring here on a daily basis so are quite grateful for feedback.

kzkvv:
I got this problem too. HikariPool is filled completely when calling transactionOutbox.flush() in a loop (total=15, active=15) and then I get the error "hikaripool-1-connection unavailable". Increasing the pool will not be a good solution and will only help for a short time. 
I tried GrannyOlli's solution and it worked for me, now the pool doesn't stay full.

amseager:
+1, looks like a blocker for using it in Spring-based apps. Made a PR with the solution presented by @GrannyOlli and @kzkvv

PR #167, opened 2021-08-11
Fix leaked db connections in SpringTransactionManager

amseager:
Fixes [#134](https://github.com/gruelbox/transaction-outbox/issues/134)

Thanks to @GrannyOlli and @kzkvv for pointing it out.

To sum up: the method [`inTransactionReturnsThrows`](https://github.com/gruelbox/transaction-outbox/blob/master/transactionoutbox-spring/src/main/java/com/gruelbox/transactionoutbox/SpringTransactionManager.java#L33) of `SpringTransactionManager` is marked as `@Transactional` but this annotation takes effect only if this method is called outside its class (e.g. [from here](https://github.com/gruelbox/transaction-outbox/blob/master/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java#L253)).

Currently, there are 4 default methods of [`TransactionManager`](https://github.com/gruelbox/transaction-outbox/blob/master/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionManager.java) which both call it too, and, in fact, from the same class (`SpringTransactionManager` inherits this interface), so `@Transactional` doesn't work in this case because of [Spring AOP restrictions](https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#aop-pointcuts). In some cases, this can cause errors (e.g. a connection can be left open after an exception occurring inside the transaction).

In this PR, I overrode those methods with the same implementations but also added `@Transactional` for them. So now, a new transaction will be created while calling them too. On the other hand, `inTransactionReturnsThrows` can be safely called inside those methods because, in this case, its `@Transactional` is ignored and there won't be extra transactions spawned (as it could be if we call interface default methods with the `super` keyword).

Another solution would be to move `inTransactionReturnsThrows` to a different class, but since it's a Spring-only related issue, I think it's better to leave it as is.

### DIFF
--- a/transactionoutbox-spring/src/main/java/com/gruelbox/transactionoutbox/SpringTransactionManager.java
+++ b/transactionoutbox-spring/src/main/java/com/gruelbox/transactionoutbox/SpringTransactionManager.java
@@ -1,5 +1,8 @@
 package com.gruelbox.transactionoutbox;
 
+import static com.gruelbox.transactionoutbox.Utils.uncheck;
+import static com.gruelbox.transactionoutbox.Utils.uncheckedly;
+
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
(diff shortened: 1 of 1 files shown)


badgerwithagun:
Great! I will take the Spring users' word for this. Looks good.

PR #178, opened 2021-09-13
Fix duration format in idempotency protection log

### LINKED ISSUE #177: Wrong duration format in idempotency protection log
The message "Expired idempotency protection on {} requests completed more than {} ago" shows incorrect duration - all values are "shifted" - e.g. "1d:00h:00m" when retention threshold is 1 hour.

amseager:
Fixes #177 
Also, uses built-in methods of `Duration` class to get days/hours/minutes/seconds instead of manual calculation


badgerwithagun:
Thanks! All merged.

PR #184, opened 2021-09-30
bugfix: DefaultPersistor should honour migrate configuration as per documentation

soandrew:
According to the documentation the `migrate()` configuration on `DefaultPersistor.builder()` should `"Disable automatic creation and migration of the outbox table"` but right now it doesn't do anything.

This PR is intended to fix that.


amseager:
Is TXNO_VERSION still generated when this config is disabled?

soandrew:
@amseager Nope, TXNO_VERSION and TXNO_OUTBOX both won't be created, which I believe is the supposed intended behaviour of that config if end users are to be managing migrations themselves.

I can update the test to check that TXNO_VERSION is not created as well?

PR #202, opened 2021-11-22
✨ replaced CGLIB with ByteBuddy

stustill:
CGLIB no longer seems to be actively maintained, with the last release being in 2019.

This change replaces CGLIB with [ByteBuddy](https://bytebuddy.net/#/).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,6 +64,7 @@
         <groupId>org.objenesis</groupId>
         <artifactId>objenesis</artifactId>
         <version>3.2</version>
+        <optional>true</optional>
       </dependency>
       <dependency>
         <groupId>com.google.code.gson</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -58,6 +58,7 @@
         <groupId>net.bytebuddy</groupId>
         <artifactId>byte-buddy</artifactId>
         <version>1.12.1</version>
+        <optional>true</optional>
       </dependency>
       <dependency>
         <groupId>org.objenesis</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -55,9 +55,9 @@
         <version>1.7.32</version>
       </dependency>
       <dependency>
-        <groupId>cglib</groupId>
-        <artifactId>cglib</artifactId>
-        <version>3.3.0</version>
+        <groupId>net.bytebuddy</groupId>
+        <artifactId>byte-buddy</artifactId>
+        <version>1.12.1</version>
       </dependency>
       <dependency>
         <groupId>org.objenesis</groupId>
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/ProxyFactory.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/ProxyFactory.java
@@ -1,0 +1,86 @@
+package com.gruelbox.transactionoutbox;
+
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+import java.util.function.BiFunction;
+import net.bytebuddy.ByteBuddy;
+import net.bytebuddy.TypeCache;
+import net.bytebuddy.TypeCache.Sort;
+import net.bytebuddy.description.modifier.Visibility;
+import net.bytebuddy.dynamic.loading.ClassLoadingStrategy;
+import net.bytebuddy.implementation.InvocationHandlerAdapter;
+import net.bytebuddy.matcher.ElementMatchers;
+import org.objenesis.Objenesis;
+import org.objenesis.ObjenesisStd;
+import org.objenesis.instantiator.ObjectInstantiator;
+
+public class ProxyFactory {
+
+  private static final Objenesis objenesis = new ObjenesisStd();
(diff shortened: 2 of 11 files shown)


stustill:
Probably need to do something to prevent lots of classes being generated, each time an outbox call is scheduled. So don't think this is quite ready.

pom.xml
```diff
@@ -55,9 +55,9 @@
         <version>1.7.32</version>
       </dependency>
       <dependency>
-        <groupId>cglib</groupId>
-        <artifactId>cglib</artifactId>
-        <version>3.3.0</version>
+        <groupId>net.bytebuddy</groupId>
+        <artifactId>byte-buddy</artifactId>
+        <version>1.12.1</version>
```
badgerwithagun, line 60: Could we make this a compile-only dependency and catch the `NoClassDefFoundError` gracefully if not present? Seems wrong for this to _need_ to be on the classpath.

badgerwithagun (changes requested):
I don't think this looks notably worse than the CGLIB implementation in terms of lots of classes being created, but I'll leave this with you to refine!

transactionoutbox-core/src/test/java/com/gruelbox/transactionoutbox/acceptance/TestMySql8.java
```diff
@@ -14,7 +14,7 @@
   @Container
   @SuppressWarnings("rawtypes")
   private static final JdbcDatabaseContainer container =
-      new MySQLContainer<>("mysql:8").withStartupTimeout(Duration.ofHours(1));
+      new MySQLContainer<>("mysql:8.0.27").withStartupTimeout(Duration.ofMinutes(5));
```
badgerwithagun, line 17: The intent here is to break the test if it fails against the latest mysql.

What's the thinking here?

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/ProxyFactory.java
```diff
@@ -0,0 +1,86 @@
...
+import net.bytebuddy.implementation.InvocationHandlerAdapter;
+import net.bytebuddy.matcher.ElementMatchers;
+import org.objenesis.Objenesis;
+import org.objenesis.ObjenesisStd;
+import org.objenesis.instantiator.ObjectInstantiator;
+
+public class ProxyFactory {
+
+  private static final Objenesis objenesis = new ObjenesisStd();
+  private static final TypeCache<Class<?>> byteBuddyCache = new TypeCache<>(Sort.WEAK);
```
badgerwithagun, line 21: Does this need to be static, based on the usage patterns you have in place?

transactionoutbox-core/src/test/java/com/gruelbox/transactionoutbox/acceptance/TestMySql8.java
stustill: I couldn't get test containers to start with 8. Seemed to work once with 8.0.27.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/ProxyFactory.java
stustill: Good point, forgot to update this after stripping it out of `Utils`.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/ProxyFactory.java
```diff
@@ -0,0 +1,86 @@
...
+                  new ByteBuddy()
+                      .subclass(clazz)
+                      .defineField("handler", InvocationHandler.class, Visibility.PUBLIC)
+                      .method(ElementMatchers.isDeclaredBy(clazz))
+                      .intercept(InvocationHandlerAdapter.toField("handler"))
+                      .make()
+                      .load(clazz.getClassLoader(), ClassLoadingStrategy.Default.INJECTION)
+                      .getLoaded());
+    } catch (NoClassDefFoundError error) {
+      throw new MissingOptionalDependencyException("net.bytebuddy", "byte-buddy");
```
badgerwithagun, line 83: Could we do the same for Objenesis?

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/ProxyFactory.java
badgerwithagun: This will fail on class init if bytebuddy isn't present.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/ProxyFactory.java
stustill: Done

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/ProxyFactory.java
stustill: Updated to catch `NoClassDefFoundError` when initialising the object, and set these things as `null`. We then check for null before we use them.

pom.xml
stustill: Done

badgerwithagun:
Hey @stustill. Would be good to get this merged in.  Are you hoping to pick it up?

PR #223, opened 2022-01-13
Permit UUID by default

### LINKED ISSUE #217: DefaultInvocationSerializer does not allow java.util.UUID
`java.util.UUID` is a commonly-used JDK class and is commonly used to model surrogate identifiers. It is therefore a good candidate to be passed as an argument to "scheduled" invocations with transaction outbox.

`DefaultInvocationSerializer` does not currently appear to allow it. There are various work-arounds, but it seems like something that should be allowed.  Do you agree? If so I will raise a PR.

tsg21:
Make `DefaultInvocationSerializer` permit `java.util.UUID`

Fixes #217

### DIFF
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultInvocationSerializer.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultInvocationSerializer.java
@@ -42,6 +42,8 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.TimeZone;
+import java.util.UUID;
+
 import lombok.Builder;
 import lombok.extern.slf4j.Slf4j;
 
@@ -54,6 +56,7 @@ import lombok.extern.slf4j.Slf4j;
  *   <li>Primitive types such as {@code int} or {@code double} or the boxed equivalents
  *   <li>{@link String}
  *   <li>{@link java.util.Date}
+ *   <li>{@link java.util.UUID}
  *   <li>The {@code java.time} classes:
  *       <ul>
  *         <li>{@link java.time.DayOfWeek}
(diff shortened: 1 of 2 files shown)


transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultInvocationSerializer.java
```diff
@@ -125,41 +128,44 @@ public Invocation deserializeInvocation(Reader reader) {
       addClassPair(boolean.class, "boolean");
       addClassPair(char.class, "char");
 
-      addClassPair(Byte.class, "Byte");
-      addClassPair(Short.class, "Short");
-      addClassPair(Integer.class, "Integer");
-      addClassPair(Long.class, "Long");
-      addClassPair(Float.class, "Float");
-      addClassPair(Float.class, "Float");
-      addClassPair(Float.class, "Float");
```
tsg21, line 134: Note the duplication here, this has been removed.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultInvocationSerializer.java
```diff
@@ -125,41 +128,44 @@ public Invocation deserializeInvocation(Reader reader) {
...
+      addClass(MonthDay.class);
+      addClass(Period.class);
+      addClass(Year.class);
+      addClass(YearMonth.class);
+      addClass(ZoneOffset.class);
+      addClass(DayOfWeek.class);
+      addClass(ChronoUnit.class);
+
+      addClass(Transaction.class);
+      addClass(TransactionContextPlaceholder.class);
```
amseager, line 160: That differs from `addClassPair(TransactionContextPlaceholder.class, "TransactionContext");`, is it ok?

tsg21: Good spot! I went through them and failed on the last line. I will undo that - I don't have an interest in accidentally altering behaviour here - the old name might be persisted.

tsg21: Fixed. Reverted to the old name for this one.

badgerwithagun:
Thanks @tsg21. Merged :)

PR #238, opened 2022-02-04
Enabled multiple java version 11, 15, 17 compatibility build

hamid646m:


badgerwithagun (dismissed):

hamid646m:
it should have failed for 17 and windows as i was not sure if the condition i added would work for docker. i included docker for windows too. now, the only thing that should fail is java 17 ( for all OS until the other PR is merged)

badgerwithagun:
@hamid646m looks like we're still getting failures.

hamid646m:
@badgerwithagun  it was missing my changes from other branch. So ubuntu-17 should definitely pass, not sure about others, i will take a look

hamid646m:
so docker is not install by default on windows and mac os and it doesn't seem easy to install ones. i will remove it for now.

PR #239, opened 2022-02-04
replaced CGLIB with ByteBuddy 

hamid646m:


pom.xml
```diff
@@ -206,22 +208,22 @@
...
-        </configuration>
-        <executions>
-          <execution>
-            <goals>
-              <goal>format</goal>
-            </goals>
-          </execution>
-        </executions>
-      </plugin>
+<!--      <plugin>-->
```
badgerwithagun, line 211: Commented out?

hamid646m, line 211: https://github.com/coveooss/fmt-maven-plugin/issues/99

not compatible with java 17 yet

badgerwithagun (changes requested):

pom.xml
hamid646m: this plugin is not compatible with java 17 https://github.com/coveooss/fmt-maven-plugin/issues/99. should i remove it?

badgerwithagun: It's possible to override behaviour for specific java versions in maven:
```
<profile>
    <id>active-on-jdk-11</id>
    <activation>
        <jdk>11</jdk>
    </activation>
</profile>
```
So you can enable this plugin for all java versions prior to 17

badgerwithagun: It's possible to override behaviour for specific java versions in maven:
```
<profile>
    <id>active-on-jdk-11</id>
    <activation>
        <jdk>11</jdk>
    </activation>
</profile>
```
So you can enable this plugin for all java versions prior to 17

hamid646m: cool, i will do that

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultInvocationSerializer.java
```diff
@@ -329,6 +338,112 @@ private String nameForClass(Class<?> clazz) {
     }
   }
 
+  static final class LocalDateTimeTypeAdapter extends TypeAdapter<LocalDateTime> {
+
+    @Override
+    public void write(JsonWriter out, LocalDateTime value) throws IOException {
+      out.value(value.format(DateTimeFormatter.ISO_LOCAL_DATE_TIME));
```
badgerwithagun, line 345: Should these have the null checks used in `UtcTypeAdapter`?

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultInvocationSerializer.java
```diff
@@ -329,6 +338,112 @@ private String nameForClass(Class<?> clazz) {
...
+      out.value(value.format(formatter));
+    }
+
+    @Override
+    public MonthDay read(JsonReader in) throws IOException {
+      return MonthDay.parse(in.nextString(), formatter);
+    }
+  }
+
+  static final class PeriodTypeAdapter extends TypeAdapter<Period> {
```
badgerwithagun, line 408: A few of these are basically identical and could just be a `ToStringTypeAdapter`

badgerwithagun (changes requested):
Looking great. Just a couple of small suggestions

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultInvocationSerializer.java
hamid646m: nor required, added a test for null

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultInvocationSerializer.java
hamid646m: but `read` method  won't be easy to implement

badgerwithagun (approved):

Issue #198, opened 2021-11-05
Question: Order of outbox entry processing


rvervaek:
I can't find any documentation about the 'ordering' of outbox entries that will be processed. 

Sometimes it's desirable that the order in which outbox entries are created (for example in case of 'events'), is the same order in which outbox entries are provided for submission (via submitter). This can be hard to achieve when running multiple instances of the same application and would require locking on database level.

Is this library able to achieve same outbox-entry-creation outbox-entry-submission ordering semantics?
If so, does this apply for running multiple instances?

Thanks in advance.

Ruben

badgerwithagun:
At the moment, no. It would be relatively trivial to achieve, but would, by design, force all tasks to be processed in single-threaded fashion. If a task failed, all work would have to queue up behind it.

What you'd end up with is a very slow, poorly designed FIFO queuing system.

That said, neither would be a major issue if all the task did was to push the work into a _real_ queuing system, like Kafka, Kinesis, RabbitMQ in ordered mode or similar, depending on your other requirements. Then the likelihood of failure/blocking would be very low and easily resolved.

So yes, I could see this working as long as performance is not a primary concern and the probability of failure is extremely low and due to some infrastructure issue rather than a code one.

Would happily take suggested PRs and I'll have a ponder about it myself.

badgerwithagun:
Also: how concurrent this could be on WRITE would depend on how strictly ordered this needs to be. If it can afford to be roughly sequential on the order of write, you could use a timestamp on write.

If it needs to be strictly ordered based on the order in which the original transactions are applied, a database sequence number/autonumber would be needed and that's going to be a bottleneck on write performance.

rvervaek:
Thanks for the explanation. My usecase is in fact a matter of sending the messages to a RabbitMQ exchange. One property of the concept of an 'event' is that it occurred on a specific time, and so my idea was to define the ordering of the messages based on that timestamp. Performance in this case is of less importance, and when processing of a specific task fails, the focus is on fixing the problem and retrying the message processing rather than making sure that all other tasks are not blocked (trying to achieve an 'eventual consistent' process). 

I will ponder about is myself as well, but at first glance I would think the introduction of a 'creationTime' property on the outboxmessage could already be enough to support an ordered submission process, based on that creationTime. 

[...]

PR #252, opened 2022-02-17
tasks ordered processing (draft)

amseager:
Plan to add the functionality described in https://github.com/gruelbox/transaction-outbox/issues/198

For tasks that need to be ordered, the main idea is to use a separate `TransactionOutbox` instance with a shorter next attempt and job flush intervals, and optionally with disabled `submitNow` functionality (so tasks are being sent only by a "flush"-job which now has a new `flushOrdered` method).
Currently, ordering is happened by the new `createTime` property in each group.
It's intended to set some value in `groupId()` of `transactionOutbox.with()` builder.

Would like to get some comments about the approach if you don't mind. 
Ofc I'll add docs/tests for it later (decided not to do it until a superficial review).

### DIFF
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
@@ -72,7 +72,18 @@ class DefaultMigrationManager {
               8,
               "Update length of invocation column on outbox for MySQL dialects only.",
               "ALTER TABLE TXNO_OUTBOX MODIFY COLUMN invocation MEDIUMTEXT",
-              Map.of(Dialect.POSTGRESQL_9, "", Dialect.H2, "")));
+              Map.of(Dialect.POSTGRESQL_9, "", Dialect.H2, "")),
+          new Migration(
+              9,
+              "Add createTime column to outbox",
+              "ALTER TABLE TXNO_OUTBOX ADD COLUMN createTime TIMESTAMP(6) NULL AFTER invocation",
+              Map.of(
+                  Dialect.POSTGRESQL_9,
+                  "ALTER TABLE TXNO_OUTBOX ADD COLUMN createTime TIMESTAMP(6)")),
+          new Migration(
+              10,
+              "Add groupId column to outbox",
+              "ALTER TABLE TXNO_OUTBOX ADD COLUMN groupId VARCHAR(250)"));
 
   static void migrate(TransactionManager transactionManager, @NotNull Dialect dialect) {
     transactionManager.inTransaction(
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultPersistor.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultPersistor.java
@@ -37,7 +37,8 @@ import lombok.extern.slf4j.Slf4j;
 public class DefaultPersistor implements Persistor {
 
   private static final String ALL_FIELDS =
-      "id, uniqueRequestId, invocation, lastAttemptTime, nextAttemptTime, attempts, blocked, processed, version";
+      "id, uniqueRequestId, invocation, lastAttemptTime, nextAttemptTime, attempts, blocked, processed, version, "
+          + "createTime, groupId";
 
   /**
    * @param writeLockTimeoutSeconds How many seconds to wait before timing out on obtaining a write
@@ -93,7 +94,7 @@ public class DefaultPersistor implements Persistor {
   public void save(Transaction tx, TransactionOutboxEntry entry)
       throws SQLException, AlreadyScheduledException {
     var insertSql =
-        "INSERT INTO " + tableName + " (" + ALL_FIELDS + ") VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)";
+        "INSERT INTO " + tableName + " (" + ALL_FIELDS + ") VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)";
     var writer = new StringWriter();
     serializer.serializeInvocation(entry.getInvocation(), writer);
     if (entry.getUniqueRequestId() == null) {
@@ -133,6 +134,9 @@ public class DefaultPersistor implements Persistor {
     stmt.setBoolean(7, entry.isBlocked());
     stmt.setBoolean(8, entry.isProcessed());
     stmt.setInt(9, entry.getVersion());
+    stmt.setTimestamp(
+        10, entry.getCreateTime() == null ? null : Timestamp.from(entry.getCreateTime()));
+    stmt.setString(11, entry.getGroupId());
   }
 
   @Override
@@ -241,8 +245,47 @@ public class DefaultPersistor implements Persistor {
                     + ALL_FIELDS
                     + " FROM "
                     + tableName
-                    + " WHERE nextAttemptTime < ? AND blocked = false AND processed = false LIMIT ?"
+                    + " WHERE nextAttemptTime < ? AND blocked = false AND processed = false and groupid is null LIMIT ?"
                     + forUpdate)) {
+      stmt.setTimestamp(1, Timestamp.from(now));
+      stmt.setInt(2, batchSize);
+      return gatherResults(batchSize, stmt);
+    }
+  }
+
+  @Override
+  public List<TransactionOutboxEntry> selectBatchOrdered(
+      final Transaction tx, final int batchSize, final Instant now) throws Exception {
+    String forUpdate = dialect.isSupportsSkipLock() ? " FOR UPDATE SKIP LOCKED" : "";
+    try (PreparedStatement stmt =
+        tx.connection()
+            .prepareStatement(
+                dialect.isSupportsWindowFunctions()
+                    ? "WITH t AS"
+                        + " ("
+                        + "   SELECT rank() over (PARTITION BY groupid ORDER BY createtime) AS rn, " + ALL_FIELDS
+                        + "   FROM " + tableName
+                        + "   WHERE processed = false"
+                        + " ) "
+                        + " SELECT " + ALL_FIELDS
+                        + " FROM t "
+                        + " WHERE rn = 1 AND nextAttemptTime < ? AND blocked = false AND groupid IS NOT null LIMIT ?"
+                        + forUpdate
+                    :
+                      "SELECT "
+                        + ALL_FIELDS
+                        + " FROM"
+                        + " ("
+                        + "   SELECT groupid AS group_id, MIN(createtime) AS min_create_time"
+                        + "   FROM " + tableName
+                        + "   WHERE processed = false"
+                        + "   GROUP BY group_id"
+                        + " ) AS t"
+                        + " JOIN " + tableName + " t1"
+                        + " ON t1.groupid = t.group_id AND t1.createtime = t.min_create_time"
+                        + " WHERE nextAttemptTime < ? AND blocked = false AND groupid IS NOT null LIMIT ?"
+                        + forUpdate
+            )) {
       stmt.setTimestamp(1, Timestamp.from(now));
       stmt.setInt(2, batchSize);
       return gatherResults(batchSize, stmt);
@@ -289,6 +332,11 @@ public class DefaultPersistor implements Persistor {
               .blocked(rs.getBoolean("blocked"))
               .processed(rs.getBoolean("processed"))
               .version(rs.getInt("version"))
+              .createTime(
+                  rs.getTimestamp("createTime") == null
+                      ? null
+                      : rs.getTimestamp("createTime").toInstant())
+              .groupId(rs.getString("groupId"))
               .build();
       log.debug("Found {}", entry);
       return entry;
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutbox.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutbox.java
@@ -88,6 +88,8 @@ public interface TransactionOutbox {
   @SuppressWarnings("UnusedReturnValue")
   boolean flush();
 
+  boolean flushOrdered();
+
   /**
    * Unblocks a blocked entry and resets the attempt count so that it will be retried again.
    * Requires an active transaction and a transaction manager that supports thread local context.
@@ -138,6 +140,7 @@ public interface TransactionOutbox {
     protected Boolean serializeMdc;
     protected Duration retentionThreshold;
     protected Boolean initializeImmediately;
+    protected Boolean submitImmediately;
 
     protected TransactionOutboxBuilder() {}
 
@@ -285,6 +288,16 @@ public interface TransactionOutbox {
     }
 
     /**
+     * @param submitImmediately If true, a task will be submitted immediately after scheduling.
+     * Defaults to true.
+     * @return Builder.
+     */
+    public TransactionOutboxBuilder submitImmediately(boolean submitImmediately) {
+      this.submitImmediately = submitImmediately;
+      return this;
+    }
+
+    /**
      * Creates and initialises the {@link TransactionOutbox}.
      *
      * @return The outbox implementation.
@@ -309,6 +322,10 @@ public interface TransactionOutbox {
      * @return Builder.
      */
     ParameterizedScheduleBuilder uniqueRequestId(String uniqueRequestId);
+
+    default ParameterizedScheduleBuilder groupId(String groupId) {
+      return this;
+    }
 
     /**
      * Equivalent to {@link TransactionOutbox#schedule(Class)}, but applying additional parameters
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java
@@ -46,6 +46,7 @@ class TransactionOutboxImpl implements TransactionOutbox {
   private final boolean serializeMdc;
   private final Validator validator;
   @NotNull private final Duration retentionThreshold;
+  @NotNull private final boolean submitImmediately;
   private final AtomicBoolean initialized = new AtomicBoolean();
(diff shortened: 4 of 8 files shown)


amseager:
There are some open questions from the original issue that are not covered atm:
- ordering by any other property (like in priority queue) - don't know if it's really needed tbh
- selecting all tasks from a group in a single job iteration but processing them in order - could be designed and made in the future, but currently, we can just use short job intervals to achieve almost the same

badgerwithagun:
Is this awaiting review @amseager? It still says "draft" in the title :)

amseager:
@badgerwithagun sorry, probably pressed "ready to review" accidentally

At the same time, could you please make a quick review of the chosen approach? The production code is more or less ready atm, but I'm still not sure if it's ok conceptually

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
```diff
@@ -72,7 +72,18 @@
               8,
               "Update length of invocation column on outbox for MySQL dialects only.",
               "ALTER TABLE TXNO_OUTBOX MODIFY COLUMN invocation MEDIUMTEXT",
-              Map.of(Dialect.POSTGRESQL_9, "", Dialect.H2, "")));
+              Map.of(Dialect.POSTGRESQL_9, "", Dialect.H2, "")),
+          new Migration(
+              9,
+              "Add createTime column to outbox",
+              "ALTER TABLE TXNO_OUTBOX ADD COLUMN createTime TIMESTAMP(6) NULL AFTER invocation",
```
badgerwithagun, line 79: Timestamps aren't guaranteed 100% reliable for ordering; it's theoretically possible to write two records at exactly the same time.

I guess the question is whether that's of particular concern.  We are watching this PR with interest here, since we may have a use for it, but our use case actually doesn't care about ordering at all; we're looking to use it to throttle processing to one task per group id at a time as a way of "rationing" our batch processing capability between customers.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
```diff
@@ -72,7 +72,18 @@
...
+              Map.of(
+                  Dialect.POSTGRESQL_9,
+                  "ALTER TABLE TXNO_OUTBOX ADD COLUMN createTime TIMESTAMP(6)")),
+          new Migration(
+              10,
+              "Add groupId column to outbox",
```
badgerwithagun, line 85: I wonder whether `partitionId` or `topicId` might be more obvious terms for what this is doing?

`topicId` probably being my favourite.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultPersistor.java
```diff
@@ -241,14 +245,53 @@ public boolean unblock(Transaction tx, String entryId) throws Exception {
...
   }
 
+  @Override
+  public List<TransactionOutboxEntry> selectBatchOrdered(
+      final Transaction tx, final int batchSize, final Instant now) throws Exception {
+    String forUpdate = dialect.isSupportsSkipLock() ? " FOR UPDATE SKIP LOCKED" : "";
+    try (PreparedStatement stmt =
+        tx.connection()
+            .prepareStatement(
+                dialect.isSupportsWindowFunctions()
```
badgerwithagun, line 263: Strictly, supports window functions _and CTEs_.  Worth having two flags here?

badgerwithagun: Or just ditch the CTE and replace it with a subquery...

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultPersistor.java
```diff
@@ -241,14 +245,53 @@ public boolean unblock(Transaction tx, String entryId) throws Exception {
...
+                    ? "WITH t AS"
```
badgerwithagun, line 264: Have you done an `EXPLAIN` on these queries on the different DBs and determined if we have suitable indexes?

I have a feeling not having `groupId` anywhere is going to make this pretty slow.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutbox.java
```diff
@@ -88,6 +88,8 @@ static TransactionOutboxBuilder builder() {
   @SuppressWarnings("UnusedReturnValue")
```
badgerwithagun, line 88: Javadoc needs updating to reflect that grouped records are ignored and these should be handled by `flushOrdered`

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutbox.java
```diff
@@ -88,6 +88,8 @@ static TransactionOutboxBuilder builder() {
...
   boolean flush();
 
+  boolean flushOrdered();
```
badgerwithagun, line 91: `flushTopics()` (if you agree with me on the naming)

badgerwithagun: Also needs Javadoc

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutbox.java
```diff
@@ -284,6 +287,16 @@ public TransactionOutboxBuilder initializeImmediately(boolean initializeImmediat
       return this;
     }
 
+    /**
+     * @param submitImmediately If true, a task will be submitted immediately after scheduling.
+     * Defaults to true.
+     * @return Builder.
+     */
+    public TransactionOutboxBuilder submitImmediately(boolean submitImmediately) {
```
badgerwithagun, line 295: At the moment, you can do `groupId(...)` without `submitImmediately(...)` and it looks like the behaviour is unexpected; because the task is submitted immediately, it ignores ordering.

Can we set `submitImmediately` to `true` automatically and refuse to allow `submitImmediately(false)` in combination with a `groupId`?

Better still: is this option required at all? I can't see that it adds anything other than a delay, if you're not using a `groupId`.  So why not just remove it?

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java
```diff
@@ -111,23 +114,34 @@ public ParameterizedScheduleBuilder with() {
...
-  private List<TransactionOutboxEntry> flush(Instant now) {
+  private List<TransactionOutboxEntry> flush(Instant now, boolean isOrdered) {
     log.debug("Flushing stale tasks");
     var batch =
         transactionManager.inTransactionReturns(
             transaction -> {
               List<TransactionOutboxEntry> result = new ArrayList<>(flushBatchSize);
-              uncheckedly(() -> persistor.selectBatch(transaction, flushBatchSize, now))
-                  .forEach(
+              uncheckedly(() -> isOrdered ?
```
badgerwithagun, line 141: You're probably going to need a standalone single-thread `ExecutorService` for this - otherwise the tasks will run in parallel, even if you submit them in order.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java
```diff
@@ -111,23 +114,34 @@ public ParameterizedScheduleBuilder with() {
   @SuppressWarnings("UnusedReturnValue")
   @Override
   public boolean flush() {
+    return flush(false);
```
badgerwithagun, line 117: Why not just do both flush types in the one call to `flush()`?

We just want to clear out all work. Why complicate it?

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java
```diff
@@ -230,16 +244,19 @@ public boolean unblock(String entryId, Object transactionContext) {
...
                   validator.validate(entry);
                   persistor.save(extracted.getTransaction(), entry);
-                  extracted
-                      .getTransaction()
-                      .addPostCommitHook(
-                          () -> {
-                            listener.scheduled(entry);
-                            submitNow(entry);
-                          });
+                  if (submitImmediately) {
```
badgerwithagun, line 251: `if (groupId != null)`?

badgerwithagun (changes requested):
.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
amseager: I don't need to have that "super-precise" reliability in my project either, but I think it may be useful for others. 

I'd probably go with a single sequence (or auto-incremented column) for a counter but use it only in tandem with the `createTime` field in order to mitigate a potential "max-value to min-value" problem almost completely.
Need to brush up MySQL 5 capabilities for all of that though

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
amseager: sure, no problem, although these 2 names are fit for kafka/MQs mostly - 99.9% chance that some of that is used with the outbox pattern, but it's theoretically possible to do smth else instead of sending messages (e.g. rest calls)

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultPersistor.java
amseager: probably yes, it's just happened that the 3 RDBMS which are used here and support WF also support CTE (I think the latter appeared to the SQL standard prior to the former)
about a subquery - IIRC I had some problems with that here (hence moved to CTE) but I'll try to do it again

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultPersistor.java
amseager: I kind of disagree about `groupId` because here we fetch all groups at once so the cardinality of this column is effectively 50/50 (filled or not), and I'm not sure will RDBMS ever use that index even after collecting proper statistics.

At the same time, I think we can add an index for `processed` (it's technically 50/50 too but obviously we'll always have much fewer unprocessed rows that processed ones while selecting, and also hints/statistics can help here). 
We actually have a composite index for `processed, blacklisted, nextAttemptTime` already which should work for a single `processed` column in a `where` clause too (at least in PG, need to check it it other supported DBs)

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutbox.java
amseager: Yeah, they will be added (I've already written some of them but haven't pushed yet)

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutbox.java
amseager: right, added it by analogy to other settings initially but now I see that it's mostly useless

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java
amseager: Didn't get about the concurrency problem tbh, could you elaborate pls?
Here, we have only 1st task per group selected in a single `flush` iteration, and it should impossible to select the 2nd one before completing the 1st because of `processed` condition.
So I thought the problem can occur only in a multi-instance environment which could be solved as it did in the existing (batched) select (with locks etc.)

About future chains - is it for a solution when we select all tasks of each group at one call instead?  I'm going to experiment with this approach.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java
amseager: The idea was to use separate `TransactionOutbox` instances for flushing ordered and non-ordered tasks because the scheduler params like a delay between iterations etc. (and flush frequency itself) are planned to be much fewer in the 1st case than in the 2nd. What do you think about this?

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java
amseager: discussed above, I'll remove `submitImmediately` and replace it with a `groupId != null` check

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
badgerwithagun: I've used a special sequence table for this in the past, to mimic the flexibility of oracle sequences.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java
badgerwithagun: Interesting. I'll ponder.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
badgerwithagun: Yeah, I don't mean to imply that `groupId`s _are_ an SQS queue or Kafka topic, only that they are an analog of these, so there's probably no point adding another term to the mix.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java
badgerwithagun: Yes, so the idea is that you select all tasks for all groups, but queue up each group separately, something like this:

```
ConcurrentMap<String, CompletableFuture<Void>> tasks = new ConcurrentMap<>();
...
persistor.selectBatchOrdered(transaction, flushBatchSize, now)
  .forEach(entry ->
    tasks.compute(
      entry.groupId(),
      (groupId, future) -> future == null
        ? CompletableFuture.runAsync(() -> processNow(entry), executor)
        : future.thenRunAsync(() -> processNow(entry), executor));
```

This ignores the presence of `Submitter`, which complicates things slightly, but it demonstrates my thinking/

The result of this is a chain of `CompletableFuture`s for each group, resulting in each task in a group running after the last task in the group completes.

makeamark1:
@amseager Do you still have local changes that you want to push? Would be great to get this pull request finished and merged.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
makeamark1: [Message groups](https://docs.aws.amazon.com/sns/latest/dg/fifo-message-grouping.html) is the terminology AWS SNS/SQS FIFOs use for the boundary within which ordering is ensured, so the term makes sense at least for those users.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutbox.java
makeamark1: @amseager Would you please be able to push those changes?

markushc:
@amseager Do you still have local changes that you want to push? Would be great to get this pull request finished and merged.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
markushc: [Message groups](https://docs.aws.amazon.com/sns/latest/dg/fifo-message-grouping.html) is the term AWS SNS/SQS FIFOs use for the boundary within which ordering is ensured (within a queue or topic), so the “group” term makes sense at least for those users. For Kafka users “partition” would be the equivalent (once again within a topic). So both group or partition seem to be reasonable choices.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutbox.java
markushc: @amseager Would you be able to push those changes?

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultMigrationManager.java
makeamark1: @amseager With the potential "max-value to min-value" problem, are you referring to a sequence wrapping around from its maximum value to its minimum value?

markushc: @amseager With the potential "max-value to min-value" problem, are you referring to a sequence wrapping around from its maximum value to its minimum value?

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutbox.java
markushc: The submit immediately option might not be required to implement ordered processing. However, it could be useful if you want to separate transaction-processing compute from task-processing compute, i.e. only process tasks using `flush` from a separate JVM. This also makes the transaction-outbox library more suitable for use with serverless compute like AWS Lambda functions, where tasks cannot be processed in background threads.

badgerwithagun:
Fixed by #588

PR #299, opened 2022-07-19
Bump javax.el 3.0.0 to jakarta.el 4.0.2 to fix CVE-2021-28170

tsg21:
The version of javax.el being pulled in by transaction-outbox has a CVE: https://www.cve.org/CVERecord?id=CVE-2021-28170

javax.el was renamed to jakarta.el as part of some earlier JDK work. The old javax.el is no longer maintained. This PR shifts transaction-outbox to the replacement library.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,14 +78,14 @@
         <version>6.2.3.Final</version>
       </dependency>
       <dependency>
-        <groupId>javax.el</groupId>
-        <artifactId>javax.el-api</artifactId>
-        <version>3.0.0</version>
+        <groupId>jakarta.el</groupId>
+        <artifactId>jakarta.el-api</artifactId>
+        <version>4.0.0</version>
       </dependency>
       <dependency>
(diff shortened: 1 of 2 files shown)


tsg21:
This isn't quite right. It doesn't work with the hibernate bean validator:
https://docs.jboss.org/hibernate/stable/validator/reference/en-US/html_single/#validator-gettingstarted-uel

Might need to bump that too.

tsg21:
It doesn't feel like bean validation (and it's EL dependency) is a core part of the transaction-outbox offering. Ideally it would be hived off into an extension like you have with the Spring/Guice/etc. integrations. I imagine that would be a backwards-incompatible change though.

Can you take javax.el out of the core dependencies and document that users would need to pull in the library themselves?

badgerwithagun:
It's purely a convenience and could easily be stripped out and replaced with a bunch of `if` conditions.

Easily done; just remove the dependency and systematically replace every use case.

Issue #305, opened 2022-08-08
ThreadPoolExecutor seems to be single thread by default.


luisalves00:
After looking at the code I think you misspelled Math.min instead of Math.max: 

```
/**
   * Schedules background worh with a {@link ThreadPoolExecutor}, sized to match {@link
   * ForkJoinPool#commonPool()} (or one thread, whichever is the larger), with a maximum queue size
   * of 16384 before work is discarded.
   *
   * @return The submitter.
   */
  static Submitter withDefaultExecutor() {
    // JDK bug means this warning can't be fixed
    //noinspection Convert2Diamond
    return withExecutor(
        new ThreadPoolExecutor(
            1,
            Math.min(1, ForkJoinPool.commonPool().getParallelism()),
            0L,
            TimeUnit.MILLISECONDS,
            new ArrayBlockingQueue<Runnable>(16384)));
  }
```
Can you please confirm.

badgerwithagun:
Yup, that's a typo. I don't think many people are actually using that executor.

Will happily take a pull request.

luisalves00:
Done. When do you plan to do a release?

badgerwithagun:
Very shortly. A day or two.

PR #306, opened 2022-08-09
#305: Calculate the max and not the min between 1 and ForkJoinPool.co…

luisalves00:
Switched Math.min to Math.max to take full capabilities of the available CPUs when using the default thread pool


badgerwithagun:
Thanks @luisalves00 !

PR #320, opened 2022-08-31
Feature : Quarkus integration

RomainWilbert:
I work with Quarkus so I created a module to make transaction-outbox work with cdi-api & transaction-api.

Should work with other implentations.

I have made basic tests, feel free to add some if needed !

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,12 +30,12 @@ jobs:
           sed -i "s_\(<version>\)[^<]*_\1${revision}_g" transactionoutbox-guice/README.md
           sed -i "s_\(<version>\)[^<]*_\1${revision}_g" transactionoutbox-jooq/README.md
           sed -i "s_\(<version>\)[^<]*_\1${revision}_g" transactionoutbox-spring/README.md
-          sed -i "s_\(<version>\)[^<]*_\1${revision}_g" transactionoutbox-cdi/README.md
+          sed -i "s_\(<version>\)[^<]*_\1${revision}_g" transactionoutbox-quarkus/README.md
           sed -i "s_\(implementation 'com.gruelbox:transactionoutbox-core:\)[^']*_\1${revision}_g" README.md
           sed -i "s_\(implementation 'com.gruelbox:transactionoutbox-guice:\)[^']*_\1${revision}_g" transactionoutbox-guice/README.md
           sed -i "s_\(implementation 'com.gruelbox:transactionoutbox-jooq:\)[^']*_\1${revision}_g" transactionoutbox-jooq/README.md
           sed -i "s_\(implementation 'com.gruelbox:transactionoutbox-spring:\)[^']*_\1${revision}_g" transactionoutbox-spring/README.md
-          sed -i "s_\(implementation 'com.gruelbox:transactionoutbox-cdi:\)[^']*_\1${revision}_g" transactionoutbox-cdi/README.md
+          sed -i "s_\(implementation 'com.gruelbox:transactionoutbox-quarkus:\)[^']*_\1${revision}_g" transactionoutbox-quarkus/README.md
         env:
           GITHUB_TOKEN: ${{ github.token }}
           SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
@@ -57,12 +57,12 @@ jobs:
           sed -i "s_\(<version>\)[^<]*_\1${revision}_g" transactionoutbox-guice/README.md
           sed -i "s_\(<version>\)[^<]*_\1${revision}_g" transactionoutbox-jooq/README.md
           sed -i "s_\(<version>\)[^<]*_\1${revision}_g" transactionoutbox-spring/README.md
-          sed -i "s_\(<version>\)[^<]*_\1${revision}_g" transactionoutbox-cdi/README.md
+          sed -i "s_\(<version>\)[^<]*_\1${revision}_g" transactionoutbox-quarkus/README.md
           sed -i "s_\(implementation 'com.gruelbox:transactionoutbox-core:\)[^']*_\1${revision}_g" README.md
           sed -i "s_\(implementation 'com.gruelbox:transactionoutbox-guice:\)[^']*_\1${revision}_g" transactionoutbox-guice/README.md
           sed -i "s_\(implementation 'com.gruelbox:transactionoutbox-jooq:\)[^']*_\1${revision}_g" transactionoutbox-jooq/README.md
           sed -i "s_\(implementation 'com.gruelbox:transactionoutbox-spring:\)[^']*_\1${revision}_g" transactionoutbox-spring/README.md
-          sed -i "s_\(implementation 'com.gruelbox:transactionoutbox-cdi:\)[^']*_\1${revision}_g" transactionoutbox-cdi/README.md
+          sed -i "s_\(implementation 'com.gruelbox:transactionoutbox-quarkus:\)[^']*_\1${revision}_g" transactionoutbox-quarkus/README.md
 
       - name: Create version update pull request
         uses: gruelbox/create-pull-request@v2
--- a/pom.xml
+++ b/pom.xml
@@ -46,7 +46,7 @@
     <module>transactionoutbox-spring</module>
     <module>transactionoutbox-guice</module>
     <module>transactionoutbox-jooq</module>
-    <module>transactionoutbox-cdi</module>
+    <module>transactionoutbox-quarkus</module>
   </modules>
 
   <dependencyManagement>
--- a/transactionoutbox-quarkus/pom.xml
+++ b/transactionoutbox-quarkus/pom.xml
@@ -9,10 +9,10 @@
 	</parent>
 	<modelVersion>4.0.0</modelVersion>
 
-	<name>Transaction Outbox CDI</name>
+	<name>Transaction Outbox Quarkus</name>
 	<packaging>jar</packaging>
 	<artifactId>transactionoutbox-quarkus</artifactId>
-	<description>A safe implementation of the transactional outbox pattern for Java (CDI extension library)
+	<description>A safe implementation of the transactional outbox pattern for Java (Quarkus extension library)
   </description>
 
 	<properties>


badgerwithagun:
Thanks @RomainWilbert, and welcome!

I'll give it a quick review now.

transactionoutbox-cdi/src/main/java/com/gruelbox/transactionoutbox/CdiTransactionManager.java
```diff
@@ -0,0 +1,174 @@
...
+      {
+         this.datasource = datasource;
+      }
+
+      @Override
+      public final Connection connection()
+      {
+         try
+         {
+            return datasource.getConnection();
```
badgerwithagun, line 79: I'm a little unclear on this: By default, I'd expect a `DataSource` to return a new connection every time (or at least an available connection from a pool) whereas here it looks like you're assuming it only returns the active connection in the current transaction.

In Quarkus, is the injectable `DataSource` a proxy that ties into the current active transaction?

badgerwithagun: If I remember correctly from previous roles working with `TransactionSynchronizationRegistry`, you can get the currently active connection from there...

transactionoutbox-cdi/src/main/java/com/gruelbox/transactionoutbox/CdiInstanciator.java
```diff
@@ -0,0 +1,15 @@
...
+import javax.enterprise.inject.spi.CDI;
+
+@ApplicationScoped
+public class CdiInstanciator extends AbstractFullyQualifiedNameInstantiator
+{
+
+   @Override
+   protected Object createInstance(Class<?> clazz)
+   {
+      return CDI.current().select(clazz).get();
```
badgerwithagun, line 13: Is there any situation where you might not want to use `CDI.current()`?

badgerwithagun:
Added some small thoughts and a couple of questions, but this looks great!

Thank you again for contributing!

transactionoutbox-cdi/src/main/java/com/gruelbox/transactionoutbox/CdiTransactionManager.java
```diff
@@ -0,0 +1,174 @@
...
+import javax.transaction.TransactionSynchronizationRegistry;
+import javax.transaction.Transactional;
+import javax.transaction.Transactional.TxType;
+
+
+public class CdiTransactionManager implements ThreadLocalContextTransactionManager
+{
+   private final CdiTransaction transactionInstance;
+
+   private TransactionSynchronizationRegistry tsr;
```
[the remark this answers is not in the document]
RomainWilbert: I inspired myself from the Spring implementation which makes the transaction final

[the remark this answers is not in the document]
badgerwithagun: I mean the `tsr`. Expect that should be final too.

transactionoutbox-cdi/src/main/java/com/gruelbox/transactionoutbox/CdiTransactionManager.java
RomainWilbert: Can't find another way to get the active connection using only API methods

badgerwithagun: As long as it is definitely the active connection you get here, all good

transactionoutbox-cdi/src/main/java/com/gruelbox/transactionoutbox/CdiInstanciator.java
badgerwithagun: OK,so it sounds like you could provide the option to do:

```
@Produces
public CdiInstantiator cdiInstantiator(BeanManager bm) {
  return CdiInstantiator.create(bm);
}
```
And thus avoid the need to use the static, but that could easily be added later so no need to worry about it for now.

RomainWilbert:
Did you take a look at the tests ?

RomainWilbert:
Hi @badgerwithagun !

I am a bit confused regarding my CDI Transaction implementation. I observed that : 
- I inspired myself from the Spring implementation, but I can't figure why the Transaction is created once and for all within the Transaction manager. There should be one transaction for each proxy call ? 
- It seems that the Datasource.getConnection() method returns a different connection on each call...and connections are never closed after proxy calls are done. I reach my connection pool size quite fast (2 new connections for each call).

Thank you for your help

RomainWilbert:
OK i figured out the issue. I had to use CDI instanciation mecanism in order to make the CDITransactionManager aware of any opened transaction. All seems OK and connections are closed at the end of the process

transactionoutbox-cdi/README.md
```diff
@@ -65,13 +65,13 @@ Create your `TransactionOutbox` as a bean:
 @Produces
 public TransactionManager cdiTransactionManager(DataSource datasource, TransactionSynchronizationRegistry tsr) throws SQLException
 {
-   return new CdiTransactionManager(datasource, tsr);
+   return CdiTransactionManager.create(datasource, tsr);
```
badgerwithagun, line 68: Looks like this might be out of date.

transactionoutbox-cdi/src/main/java/com/gruelbox/transactionoutbox/CdiTransactionManager.java
```diff
@@ -2,173 +2,150 @@
...
+    }
+
+    return work.doWork(transactionInstance);
+  }
+
+  private final class CdiTransaction implements Transaction {
+
+    public final Connection connection() {
+      try {
+        return datasource.getConnection();
```
badgerwithagun, line 71: This is still not the connection used by the currently active transaction. It's requesting a new connection from the pool.

And there is no guarantee that a Spring Boot application will be using Agroal as the connection pool.

I've had a bit of a read and here's the answer: https://stackoverflow.com/questions/9642643/datasourceutils-getconnection-vs-datasource-getconnection

Just inject the normal `DataSource` and use `DataSourceUtils.getConnection(dataSource)`. That will return the currently active transaction's connection if one is active, or a fresh connection if it isn't.

To be sure that this definitely never creates a new connection, call `TransactionSynchronizationManaher.isActualTransactionActive()` first, and throw `IllegalStateException` if it returns false (a transaction should definitely be active at this point!)

RomainWilbert: I can guarantee you that with this implementation and Quarkus, this gets the current active connection.

DataSourceUtils.getConnection(dataSource) is Spring dependent so I cannot / don't want to use it

badgerwithagun: OK.

By the sound of things from this: https://quarkus.io/guides/datasource - what you're describing is not just Quarkus, but Quarkus with Agroal as the chosen connection pool _and_ Naranya for transaction management. It sounds like this is the default though.

Sounds like this is really a Quarkus integration rather than a "bare" CDI implementation, so I'll be happy with it as-is if the module is renamed to `transactionoutbox-quarkus` and `CdiTransactionManager` is renamed to `QuarkusTransactionManager`.  `CdiInstantiator` is named sensibly.

RomainWilbert: ok I renamed the module

.github/workflows/release.yml
```diff
@@ -30,10 +30,12 @@ jobs:
           sed -i "s_\(<version>\)[^<]*_\1${revision}_g" transactionoutbox-guice/README.md
           sed -i "s_\(<version>\)[^<]*_\1${revision}_g" transactionoutbox-jooq/README.md
           sed -i "s_\(<version>\)[^<]*_\1${revision}_g" transactionoutbox-spring/README.md
+          sed -i "s_\(<version>\)[^<]*_\1${revision}_g" transactionoutbox-cdi/README.md
```
badgerwithagun, line 33: Needs renaming in this file too

pom.xml
```diff
@@ -46,6 +46,7 @@
     <module>transactionoutbox-spring</module>
     <module>transactionoutbox-guice</module>
     <module>transactionoutbox-jooq</module>
+    <module>transactionoutbox-cdi</module>
```
badgerwithagun, line 49: Rename?

transactionoutbox-quarkus/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<parent>
+		<artifactId>transactionoutbox-parent</artifactId>
+		<groupId>com.gruelbox</groupId>
+		<version>${revision}</version>
+	</parent>
+	<modelVersion>4.0.0</modelVersion>
+
+	<name>Transaction Outbox CDI</name>
```
badgerwithagun, line 12: Rename?

badgerwithagun (changes requested):

badgerwithagun (approved):

RomainWilbert:
Maybe something is not rightly implemented with the connection because I have errors in logs
`[io.agr.pool] (executor-thread-0) Datasource '<default>': Closing open connection prior to commit`

RomainWilbert:
We are benchmarking the module and there are issues with performance and connections lifecycle. This implementation was inspired by the Spring one, but obviously it is not the right angle. I will keep posting updates here

badgerwithagun:
Any luck @RomainWilbert ?

RomainWilbert:
False alert. We did some real time debug to ensure everything was OK on that part. Our issues came from elsewhere : a hardcoded timeout on a request in a DAO...
The warning Closing open connection prior to commit comes from the connection wrapper, there is a reaper that checks if connection & statements are closed before commit, and it is called before the afterCompletion event...one way to prevent the warnings is to return the connection instead of the wrapper in the TransactionManager : but required to be careful on statement & connection closing as the reaper secure closing is not called anymore before commit.
` public final Connection connection() {
      try {
        ConnectionWrapper connectionWrapper = (ConnectionWrapper) datasource.getConnection();
        Connection connection = connectionWrapper.getHandler().rawConnection();
        log.trace("Connection : " + connection);
        connectionWrapper.close();
        return connection;
      } catch (SQLException e) {
        throw new RuntimeException(e);
      }
    }`

PR #367, opened 2022-12-15
Fix postgres dialect delete expired

kzkvv:
The DELETE SQL of POSTGRESQL_9 dialect ends up slowly executing and producing excessive CPU load.

Quoted from [PostgreSQL - System Columns](http://www.postgresql.org/docs/current/static/ddl-system-columns.html):

> ctid: The physical location of the row version within its table. Note that although the ctid can be used to locate the row version very quickly, a row's ctid will change each time it is updated or moved by VACUUM FULL. Therefore ctid is useless as a long-term row identifier. The OID, or even better a user-defined serial number, should be used to identify logical rows.

The original uses "ctid" system value, but the modified uses the "id" primary key column. Here are explain plans of the original query and of modified query, showing that the modified one doesn't have that unexplainably high cost.

original:
```
explain DELETE FROM TXNO_OUTBOX WHERE ctid IN (SELECT ctid FROM TXNO_OUTBOX WHERE nextAttemptTime < now() AND processed = true AND blocked = false LIMIT 100);                                                        
--------------------------------------------------------------------------------------------------------------------------
Delete on txno_outbox  (cost=0.13..750420.75 rows=1 width=36)
  ->  Nested Loop Semi Join  (cost=0.13..750420.75 rows=1 width=36)
        Join Filter: (txno_outbox.ctid = "ANY_subquery".ctid)
        ->  Seq Scan on txno_outbox  (cost=0.00..750419.00 rows=1 width=6)
        ->  Subquery Scan on "ANY_subquery"  (cost=0.13..1.74 rows=1 width=36)
              ->  Limit  (cost=0.13..1.73 rows=1 width=6)
                    ->  Index Scan using ix_txno_outbox_1 on txno_outbox txno_outbox_1  (cost=0.13..1.73 rows=1 width=6)
                          Index Cond: ((processed = true) AND (blocked = false) AND (nextattempttime < now()))
                          Filter: (processed AND (NOT blocked))
```
modified:
```
explain DELETE FROM TXNO_OUTBOX WHERE id IN (SELECT id FROM TXNO_OUTBOX WHERE nextAttemptTime < now() AND processed = true AND blocked = false LIMIT 100);                                                           
---------------------------------------------------------------------------------------------------------------------------------
Delete on txno_outbox  (cost=2.98..5.21 rows=1 width=67)
   ->  Nested Loop  (cost=2.98..5.21 rows=1 width=67)
         ->  HashAggregate  (cost=1.74..1.75 rows=1 width=98)
               Group Key: ("ANY_subquery".id)::text
               ->  Subquery Scan on "ANY_subquery"  (cost=0.13..1.74 rows=1 width=98)
                     ->  Limit  (cost=0.13..1.73 rows=1 width=37)
                           ->  Index Scan using ix_txno_outbox_1 on txno_outbox txno_outbox_1  (cost=0.13..1.73 rows=1 width=37)
                                 Index Cond: ((processed = true) AND (blocked = false) AND (nextattempttime < now()))
                                 Filter: (processed AND (NOT blocked))
         ->  Bitmap Heap Scan on txno_outbox  (cost=1.23..2.35 rows=1 width=43)
               Recheck Cond: ((id)::text = ("ANY_subquery".id)::text)
               ->  Bitmap Index Scan on txno_outbox_pkey  (cost=0.00..1.23 rows=1 width=0)
                     Index Cond: ((id)::text = ("ANY_subquery".id)::text)
```

The modification is only "ctid" --> "id". Functionally, the query is equivalent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -477,6 +477,8 @@ TransactionOutbox outbox = TransactionOutbox.builder()
         .dialect(Dialect.POSTGRESQL_9)
         // Override the table name (defaults to "TXNO_OUTBOX")
         .tableName("transactionOutbox")
+        // Override the column name (defaults to "id")
+        .columnName("id")
         // Shorten the time we will wait for write locks (defaults to 2)
         .writeLockTimeoutSeconds(1)
         // Disable automatic creation and migration of the outbox table, forcing the application to manage
--- a/README.md
+++ b/README.md
@@ -477,8 +477,6 @@ TransactionOutbox outbox = TransactionOutbox.builder()
         .dialect(Dialect.POSTGRESQL_9)
         // Override the table name (defaults to "TXNO_OUTBOX")
         .tableName("transactionOutbox")
-        // Override the column name (defaults to "id")
-        .columnName("id")
         // Shorten the time we will wait for write locks (defaults to 2)
         .writeLockTimeoutSeconds(1)
         // Disable automatic creation and migration of the outbox table, forcing the application to manage
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultPersistor.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultPersistor.java
@@ -3,7 +3,13 @@ package com.gruelbox.transactionoutbox;
 import java.io.IOException;
 import java.io.Reader;
 import java.io.StringWriter;
-import java.sql.*;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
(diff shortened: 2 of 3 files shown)


badgerwithagun:
Thanks @kzkvv !

The query change looks fine.

Could you remove the parameterisation though? It'll cause problems with schema management and it seems odd to parameterise that one column and not the others.

I'm quite happy if this change is simply a one-liner, `ctid` --> `id`.

kzkvv:
Done! Only the postgres DELETE query has been updated, other changes have been reverted

badgerwithagun:
Merged! Thanks again.

PR #424, opened 2023-05-03
Ensure MDC is included when writing logs around processing

### LINKED ISSUE #423: Entry failure logging omits MDC context
Transaction Outbox helpfully supports serialising logging MDC and passing this to the thread that performs the processing. This can help immensely with tracing in environments that have a lot of concurrent transactions.

When a transaction outbox entry fails, it writes a log message along the lines of:
```
Temporarily failed to process entry e512a3e7-6c7c-4dd2-bfec-f11765deab51 : service.method("arg1", "arg2") [e512a3e7-6c7c-4dd2-bfec-f11765deab51]
```

Unfortunately, this logging is written without the MDC context for that transaction outbox entry. This can make it difficult to relate the error to the processing that preceded it, and diminishes the value of the MDC serialisation feature. This seems like a bug - it would surely be preferable for the exception to be logged with the invocation's MDC?

This issue appears to be because the MDC entry/exit happens in `Invocation.invoke`, but the error handling is outside that, in `TransactionOutboxImpl.processNow`. I think this could be fixed by pulling the MDC entry/exit logic up to wrap the entire contents of 'processNow'.

I may attempt a PR, but wanted to raise this nonetheless.

tsg21:
Fixes #423

### DIFF
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java
@@ -259,39 +259,41 @@ class TransactionOutboxImpl implements TransactionOutbox, Validatable {
   @Override
   @SuppressWarnings("WeakerAccess")
   public void processNow(TransactionOutboxEntry entry) {
-    try {
-      initialize();
-      var success =
-          transactionManager.inTransactionReturnsThrows(
-              transaction -> {
-                if (!persistor.lock(transaction, entry)) {
-                  return false;
-                }
-                log.info("Processing {}", entry.description());
-                invoke(entry, transaction);
-                if (entry.getUniqueRequestId() == null) {
-                  persistor.delete(transaction, entry);
-                } else {
-                  log.debug(
-                      "Deferring deletion of {} by {}", entry.description(), retentionThreshold);
-                  entry.setProcessed(true);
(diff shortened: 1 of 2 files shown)


tsg21:
I've looked at the existing MDC test and cannot easily see how to adapt it to cover this case. We'd have to intercept the logging in some way I think.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java
```diff
@@ -259,39 +259,41 @@ private void submitNow(TransactionOutboxEntry entry) {
...
-                }
-                return true;
-              });
-      if (success) {
-        log.info("Processed {}", entry.description());
-        listener.success(entry);
-      } else {
-        log.debug("Skipped task {} - may be locked or already processed", entry.getId());
+    entry.getInvocation().withinMDC(() -> {
+      try {
```
tsg21, line 263: Mostly whitespace changes. Suggest comparing with whitespace ignored.

badgerwithagun: Could you run a full maven build? It will standardise the formatting.

badgerwithagun (approved):

badgerwithagun:
Thanks! Merged

PR #484, opened 2023-09-06
Refactored the Dialect to use an interface to generate the DB specific SQL.

wynan:
Plan to add SQL Server support and this seemed like the time to refactor the Dialect into a proper interface. No functional changes.


badgerwithagun:
This looks sensible. However, it's quite a compatibility break. This could be avoided to a degree (if we ignore binary compatibility) by:

1. Just change `Dialect` into an interface
2. Include the original implementations as inner classes of `Dialect`
3. Add `public static final` properties exposing instances of those inner classes to replace the enum properties

e.g.

```
public interface Dialect {

  static final H2 = new H2Impl();

  static final class H2Impl implements Dialect {
  }

}
```

transactionoutbox-core/src/test/java/com/gruelbox/transactionoutbox/AbstractDefaultPersistorTest.java
```diff
@@ -33,7 +33,7 @@ abstract class AbstractDefaultPersistorTest {
 
   protected abstract DefaultPersistor persistor();
 
-  protected abstract Dialect dialect();
```
badgerwithagun, line 36: There should be no reason for this class and most of the other test classes to change.

wynan:
Will do. I was trying to keep Dialect an Enum and create a new concept to encapsulate the SQL differences but I'll go ahead and change Dialect to an interface.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DialectH2Impl.java
```diff
@@ -0,0 +1,7 @@
+package com.gruelbox.transactionoutbox;
+
+import lombok.EqualsAndHashCode;
+
+/** Dialect SQL implementation for H2. */
+@EqualsAndHashCode
+public class DialectH2Impl extends DialectMySQL5Impl {}
```
badgerwithagun, line 7: The use of inheritance here fails the "is a" test.

Since `DialectMySQQL5Impl` is a public class, it forms part of the API of the library, so we have to be careful here.

I'd be much happier if all the `Dialect` implementations were `final`, and composition/forwarding were used for code sharing. Happier still if the classes were also package-private, so they had no API footprint. There's no need for them to be public, so let's not expose them.

wynan: Ok, took another stab at this. Implementations are now package private and final. Added new focused base implementations thinking it was easier to understand what functionality was supported but let me know if you were thinking to just use one of the existing implementations.

wynan:
@badgerwithagun Let me know if you see anything else you want updated or want further changes. I feel like the migration code could be refactored as well but wanted to keep this change tight and small.

PR #492, opened 2023-09-24
Support and run acceptance tests for PostgreSQL versions 11-16

mkjensen:
At this time, PostgreSQL versions 11-16 are officially supported according to https://www.postgresql.org/support/versioning/

Remove tests using versions 9 and 10. Add tests using versions 13-16.

(I wanted to contribute something and I thought this could be a start.)


mkjensen:
I noticed that `com.gruelbox.transactionoutbox.Dialect` has a `POSTGRESQL_9` constant, but I guess there is no need to change that.

badgerwithagun (approved):

PR #540, opened 2023-12-22
Provide ability to check status of a specific entry

### LINKED ISSUE #364: Bubble up the error from within the scheduled rest request
Hello, I want to throw the error from my scheduled request when making a rest call. I can see it in the logs, but it returns a 200 as the commit overrides it. It is nice that it will get retried before getting blocked, but I need to let the browser know how the request went.

Is there an easy way to achieve this?

Thanks!

badgerwithagun:
Fixes #364


badgerwithagun:
Needs Javadoc update

PR #592, opened 2024-03-19
Refactored queries to make supporting SQL Server Limit and Skip Logic easier.

wynan:
Looking to support SQL Server in the future and noticed changes were required for Limit and Skip Locking.

Some example SQL Server queries to support this:
Select Batch
SELECT TOP ({{batchSize}}) {{allFields}} FROM {{table}} WITH (UPDLOCK, ROWLOCK, READPAST) WHERE nextAttemptTime < ? AND blocked = 0 AND processed = 0
Select Version
SELECT version FROM TXNO_VERSION WITH (UPDLOCK, ROWLOCK, READPAST)
Delete Expired
DELETE TOP ({{batchSize}}) FROM {{table}} WHERE nextAttemptTime < ? AND processed = 1 AND blocked = 0

Among others.

This would be the first of a series of changes to support SQL Server.


badgerwithagun:
Thanks @wynan!

This looks pretty sensible. Though I suspect at this point that there's little point keeping the `DefaultDialect`; we may as well just make these individual classes. Maybe a little bit of inheritance for different versions of MySQL...

I'd probably consider pre-preparing the SQL strings too (e.g. interpolating `ALL_FIELDS`) on construction rather than doing it every time.

I can't see any reason to not approve this PR though assuming all the tests pass.

wynan:
Totally agree about the DefaultDialect and the pre-preparing the SQL strings as well. Wanted to be careful about making too drastic a change too quickly. I'll look at those in future PRs.

wynan:
Just wanted to confirm you don't require any changes for this PR and the comments can be addressed in future PRs?

badgerwithagun:
Thanks for reminder. Last checks now

PR #633, opened 2024-05-03
Pulled Select Next Sequence into Dialect in preparation for SQL Server.

wynan:
Needed to allow using a query similar to, "SELECT seq FROM TXNO_SEQUENCE WITH (UPDLOCK, ROWLOCK, READPAST) WHERE topic = ?" , for SQL Server when selecting the next sequence numbers.


wynan:
No rush but any concerns with this PR?

badgerwithagun (approved):
A

PR #641, opened 2024-05-16
Add MS SQL Server Support

wynan:


transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/Dialect.java
```diff
@@ -131,4 +133,64 @@ public interface Dialect {
                 }
               })
           .build();
+
+  Dialect MS_SQL_SERVER =
```
badgerwithagun, line 137: Given that this is a completely new implementation it would be nicer to just start from the latest migration id and create everything in the start state in one go, rather than replicating the full upgrade chain from when this library was first created...

transactionoutbox-acceptance/src/test/java/com/gruelbox/transactionoutbox/acceptance/persistor/TestDefaultPersistorMSSqlServer2017.java
```diff
@@ -0,0 +1,47 @@
+package com.gruelbox.transactionoutbox.acceptance.persistor;
```
badgerwithagun, line 1: Are all these SQL server implementations different enough to justify the tests?

Don't mind really, just worth checking

pom.xml
```diff
@@ -174,6 +174,18 @@
...
+      </dependency>
+      <dependency>
+        <groupId>com.microsoft.sqlserver</groupId>
+        <artifactId>mssql-jdbc</artifactId>
+        <version>12.6.1.jre11</version>
+        <scope>test</scope>
+      </dependency>
     </dependencies>
   </dependencyManagement>
   <build>
```
badgerwithagun, line 191: We should push all these MS SQL tests out to their own CI builds.  Could you create an additional profile in `pom.xml` for this? Example for postgres
```
    <profile>
      <id>only-postgres-tests</id>
      <build>
        <plugins>
          <plugin>
            <artifactId>maven-surefire-plugin</artifactId>
            <configuration>
              <includes>
                <include>**/*Postgres*.java</include>
              </includes>
            </configuration>
          </plugin>
        </plugins>
      </build>
    </profile>
```

And then add it to the build matrix here: https://github.com/gruelbox/transaction-outbox/blob/master/.github/workflows/pull_request.yml#L33

THe idea is that the tests all use H2 only unless one or more of these profiles are enabled.

badgerwithagun (changes requested):
A few small tweaks

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/Dialect.java
wynan: Should be addressed.

pom.xml
wynan: Should be addressed but let me know if I missed something.

transactionoutbox-acceptance/src/test/java/com/gruelbox/transactionoutbox/acceptance/persistor/TestDefaultPersistorMSSqlServer2017.java
wynan: It is a good question. The SQL used should be supported by all versions of the database. This was my way of testing that that was true and that hopefully any change, either in the library or in the databases, would be caught. I don't have a problem removing the other instances, though, just let me know your preference.

wynan:
Let me know if there is anything else you want me to address.

wynan:
No rush but if there is anything else you want me to address let me know.

badgerwithagun:
Looks great! Thanks.

badgerwithagun (approved):

Issue #677, opened 2024-07-27
Provide an interface for generating sequences


VladislavBakshanskij:
We have value generation going through Redis. Is it possible to add an interface so that sequence generation can be done in different ways?

badgerwithagun:
Yep, absolutely. Checking your PR now.

PR #678, opened 2024-07-28
Provide an interface for generating sequences #677

VladislavBakshanskij:


VladislavBakshanskij:
@badgerwithagun can u review?

VladislavBakshanskij:
@badgerwithagun can u review?

VladislavBakshanskij:
@badgerwithagun can u review?

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultPersistor.java
```diff
@@ -56,6 +56,12 @@ public class DefaultPersistor implements Persistor, Validatable {
   @SuppressWarnings("JavaDoc")
   private final Dialect dialect;
 
+  /**
+   * @param sequenceGenerator The sequence generator used for ordered tasks. Required
+   */
+  @SuppressWarnings("JavaDoc")
+  private final SequenceGenerator sequenceGenerator;
```
badgerwithagun, line 63: On its own, this is a compatibility break; it requires all existing users to specify `DefaultSequenceGenerator` when creating a `DefaultPersistor`.  I suggest changing this class to have a constructor annotated with `@Builder`; then you can accept `null` for `sequenceGenerator` and replace it with a `DefaultSequenceGenerator`.

transactionoutbox-acceptance/src/test/java/com/gruelbox/transactionoutbox/acceptance/TestRequestSerialization.java
```diff
@@ -30,6 +30,9 @@ final void workAlwaysSerialized() throws Exception {
             .persistor(
                 DefaultPersistor.builder()
                     .dialect(connectionDetails().dialect())
+                    .sequenceGenerator(DefaultSequenceGenerator.builder()
```
badgerwithagun, line 33: This shouldn't be necessary (see comments on `DefaultPersistor`)

transactionoutbox-acceptance/src/test/java/com/gruelbox/transactionoutbox/acceptance/TestComplexConfigurationExample.java
```diff
@@ -47,7 +48,8 @@ void test() {
                 DefaultPersistor.builder()
                     // Selecting the right SQL dialect ensures that features such as SKIP LOCKED are
                     // used correctly.
-                    .dialect(Dialect.POSTGRESQL_9)
+                    .dialect(dialect)
+                    .sequenceGenerator(DefaultSequenceGenerator.builder().dialect(dialect).build())
```
badgerwithagun, line 52: This shouldn't be necessary (see comments on `DefaultPersistor`)

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/Persistor.java
```diff
@@ -20,7 +20,10 @@ public interface Persistor {
    * @return The persistor.
    */
   static DefaultPersistor forDialect(Dialect dialect) {
-    return DefaultPersistor.builder().dialect(dialect).build();
+    return DefaultPersistor.builder()
+        .dialect(dialect)
+        .sequenceGenerator(DefaultSequenceGenerator.builder().dialect(dialect).build())
```
badgerwithagun, line 25: This shouldn't be necessary (see comments on `DefaultPersistor`)

badgerwithagun (changes requested):
One tweak needed to avoid any compatibility issues, but otherwise this looks fine.

I must caution you though: if you're planning on generating sequence numbers from outside the main DB transaction, you're going to get some unexpected timing differences.

There are several combinations to consider, but here's one:

1. Thread A asks for a sequence number and gets 1
1. Thread B asks for a sequence number and gets 2
1. Thread B writes its outbox record (sequence 2) and commits
1. `flush()` is called
1. (2) is picked up and processed
1. Thread A writes its outbox record (sequence 1) and commits
1. `flush()` is called
1. (1) is picked up and processed

The order of processing was actually 2, 1 instead of 1, 2.

So that's a bit potentially confusing, but another way of looking at this problem is that:

* `DefaultSequenceGenerator` ensures that the order of commit of the source transactions align with the order of the outgoing messages.
* A Redis implementation could only ensure that the order of the outgoing messages aligns with the _point in time that the sequence number was requested_.

VladislavBakshanskij:
@badgerwithagun I came back late of course with the fix of the comments, but I still corrected them. I ask you to do a second review

transactionoutbox-acceptance/src/test/java/com/gruelbox/transactionoutbox/acceptance/TestComplexConfigurationExample.java
VladislavBakshanskij: fixed

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultPersistor.java
VladislavBakshanskij: fixed

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/Persistor.java
VladislavBakshanskij: fixed

VladislavBakshanskij:
@badgerwithagun can u review this PR

PR #687, opened 2024-08-12
Allow TransactionOutbox#scheduler shutdown

### LINKED ISSUE #686: TransactionOutboxImpl#scheduler can't be shutdown without stopping the JVM
Hello,

Our application is deployed as a WAR in a standalone Tomcat. The application can 'reboot' without rebooting the whole JVM.
It seems there is no way to programmatically stop `TransactionOutboxImpl#scheduler` or to provide it.
We also need to customize the scheduler pool.

Could you please offer a way to provide a custom instance of `ScheduledExecutorService` in `TransactionOutboxImpl`?

reda-alaoui:
Fix #686


reda-alaoui:
@badgerwithagun > Currently it doesn't control the lifetime of the supplied Executor; your application does. You can easily shut your executor down without needing TransactionOutboxImpl to do it - in fact it seems quite strange to me to expect it to do so. Maybe I'm missing something?

Do you consider like me that the `Submitter` should be `AutoCloseable` ? Considering that the answer is yes, I apply the following rule: any `AutoCloseable` instance B *created by* another `AutoCloseable` instance A should be closed by instance A when instance A is closed. IMO, without this, apis involving `AutoCloseable` are not consistent and are error prone.

badgerwithagun:
I don't see why the `Executorubmitter` should be autocloseable, no - it doesn't manage any resource lifetimes.

Any `Executor` it works with is created outside and thus should have its lifetime managed outside.

A good rule of thumb is that whatever creates a resource should close it. `ExecutorSubmitter` doesn't create any resources.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/Submitter.java
```diff
@@ -30,15 +28,7 @@ static Submitter withExecutor(Executor executor) {
...
-    //noinspection Convert2Diamond
-    return withExecutor(
-        new ThreadPoolExecutor(
-            1,
-            Math.max(1, ForkJoinPool.commonPool().getParallelism()),
-            0L,
-            TimeUnit.MILLISECONDS,
-            new ArrayBlockingQueue<Runnable>(16384)));
+    return ExecutorSubmitter.builder().build();
   }
```
reda-alaoui, line 32: @badgerwithagun `ExecutorSubmitter` does not create an executor, but `Submitter` does here. `AutoCloseable` was originally implemented because of this.

badgerwithagun: But you don't need to use this method.

You can just create a `TransactionOutbox` with your own. 

```
var outbox = TransactionOutbox.builder()
   .submitter(ExecutorSubmitter.builder().executor(yourExecutor).build())
   ...
   .builder();
```

badgerwithagun (changes requested):
.

PR #689, opened 2024-08-12
Remove javax.inject dependency

reda-alaoui:
* The `javax` namespace has been replaced by `jakarta`.
* `javax.inject` dependency is not declared with Maven test scope despite the fact it is used only on the test side. It leaks in consuming applications.
* `javax.inject` dependency is only used by a single Spring test that can use `@Autowired` instead

Because of all that, here is a PR that simply removes `javax.inject` dependency.


badgerwithagun:
Thanks @reda-alaoui! Looks sensible. Will merge shortly.

PR #739, opened 2024-10-10
Allow overriding of attemptFrequency and blockAfterAttempts at schedule request level

lburja:
Stores the values `attemptFrequency` and `blockAfterAttempts` in the database.

This allows us to have different retry configuration per schedule request.

PR #790, opened 2025-01-29
Allow to configure retry policy on a per-task basis

nishiol:
Based on discussion: https://github.com/gruelbox/transaction-outbox/discussions/737

Similar PR: https://github.com/gruelbox/transaction-outbox/pull/739

Allows to configure retry policy on per-task basis in programmatic way. No db migrations needed.
Usage example can be found in README.md
 
@lburja could you please also take a look at this PR?

### DIFF
--- a/README.md
+++ b/README.md
@@ -426,6 +426,48 @@ However, if you completely trust your serialized data (for example, your develop
 
 See [transaction-outbox-jackson](transactionoutbox-jackson/README.md), which uses a specially-configured Jackson `ObjectMapper` to achieve this.
 
+### Flexible retry policy
+
+The default retry policy for tasks is configured using:
+```
+TransactionOutbox outbox = TransactionOutbox.builder()
+    ...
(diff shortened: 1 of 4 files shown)


lburja:
The ideea looks interesting. The upside is that DB changes are not needed, and there is flexibility in applying different strategies. The downside is that you need to implement a specific interface.

I didn’t review in detail at the code, but from what I can tell, the idea should work

nishiol:
Yes, it's kind of tight coupling. But if this a problem, it could be avoided by creating an implementing class in the module which depends on transaction-outbox stuff. It'll look like some wrapper around your business logic class which you want to be called by transaction-outbox.

badgerwithagun:
Hey, sorry about the delay here. I've proposed a solution to cover all use cases here: https://github.com/gruelbox/transaction-outbox/discussions/737#discussioncomment-13192483

PR #822, opened 2025-03-24
Populate MDC before starting transaction so it is available for pre-transaction hooks

victokoh:


badgerwithagun:
Discussed separately with @victokoh.

Her use case is that a tenant identifier is held within the MDC. This tenant identifier is used within a Hibernate query decorator [like this](https://spring.io/blog/2022/07/31/how-to-integrate-hibernates-multitenant-feature-with-spring-data-jpa-in-a-spring-boot-application) which fetches the tenant identifier at the start of a transaction.

Hibernate only calls `resolveCurrentTenantIdentifier` at the start of a transaction, when the MDC hasn't been configured yet.  By switching these around, the MDC is now populated before the transaction starts, resolving the problem.

This doesn't affect anything as far as I can see and seems like a nice easy fix with no downsides, so approving

Issue #628, opened 2024-04-30
Versioning on invocations serialized to the database


mwkent:
Hello,

I ran into an issue with the following scenario:

1. A record gets written to the outbox table in the database.
2. The record fails to process but remains unblocked.
3. New code is deployed that will expect something different with the invocation column of a record in the database (e.g., an argument now contains field B instead of field A).
4. The old record is attempted to be processed using the new code but can no longer be deserialized due to the changes deployed in the latest code.

Based on the above scenario, I am ultimately wondering if there is any way currently to support some form of versioning on records that get written to the database and how those older records would be processed if new code was deployed.

As a side note related to this, I noticed that if deserialization of a record does fail in the `flush` call, it looks like an exception is thrown, which will halt any further processing of records in that same `flush` call. Is there a way to skip processing of records that will throw an exception in the `flush` call. Otherwise, it seems like any processing of records would be halted until the issue with the specific problem record was resolved.

badgerwithagun:
Hi @mwkent:

When we do code review, one of the things we check for is developers changing the signatures of outboxed methods and treat it like an API change.  If they need to do it they create an additional method which takes different arguments and ensure the old one remains in place until we can be sure any old records are processed.

I can't really think of a way of introducing versioning as a first-class feature without introducing schema management at the same time, which starts to make the library a lot more complicated.

Maybe a combination of rules with something like [ArchUnit](https://www.archunit.org/) and/or Sonar might provide the checks you need?

PR #851, opened 2025-04-30
Increase robustness of outbox flushing with respect to outbox entry deserialization issues

skastenholz:
Background:
- We are utilizing this framework in all our microservices for certain requests to external and internal downstream services.
- In doing so JacksonInvocationSerializer is used to serialize/deserialize invocations (in a most flexible way).

Problem:
- On rare occassions some outbox entries that were serialized cannot be deserialized successfully.
- For instance this happens when the id of multiple attributes are set to null. However, the concrete issue is not relevant in this context. Most importantly the the call of "deserializeInvocation" at JacksonInvocationSerializer (https://github.com/gruelbox/transaction-outbox/blob/master/transactionoutbox-jackson/src/main/java/com/gruelbox/transactionoutbox/jackson/JacksonInvocationSerializer.java#L46) fails with an IOException.

Analysis:
- As a result the IOException is wrapped in a RuntimeException and is populated to
  - https://github.com/gruelbox/transaction-outbox/blob/master/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java#L129C53-L129C64
  - https://github.com/gruelbox/transaction-outbox/blob/master/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java#L142
- Ultimately, the processing of outbox entries stops entirely - even in case the problems applies to a single outbox entry only as part of the batch. As a result no outbox entries are processed.
- In addition the same happens when the outbox is flushed again at a later point in time. 
- Finally, all outbox entries of the microservice are no longer processed.
- A similar observation was made in #628.

Solution:
- The approach suggested here is to handle the IOException at a later point in time at the DefaultPersistor.
- During deserialization of all relevant outbox entries the ones on which the problems applied are ignored (instead of failing the whole operation).
- In doing so deserializeable entries are processed and the processing of remaining outbox entries is not  blocked.

Open points:
- [ ] Please review the solution and provide your feedback.
- [ ] With respect to tests I would plan to extend AbstractPersistorTest, but am struggling since the InvocationSerializer is not exposed yet. Any hints would be welcome.

### DIFF
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DeserializationFailedException.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DeserializationFailedException.java
@@ -1,8 +1,0 @@
-package com.gruelbox.transactionoutbox;
-
-/** Thrown when de-serialization of an invocation fails. */
-class DeserializationFailedException extends RuntimeException {
-  DeserializationFailedException(String message, Throwable cause) {
-    super(message, cause);
-  }
-}
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DeserializationFailedException.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DeserializationFailedException.java
@@ -1,0 +1,9 @@
+package com.gruelbox.transactionoutbox;
+
+
+/** Thrown when de-serialization of an invocation fails. */
+class DeserializationFailedException extends RuntimeException {
+  DeserializationFailedException(String message, Throwable cause) {
+    super(message, cause);
+  }
+}
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DeserializationFailedException.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DeserializationFailedException.java
@@ -1,6 +1,5 @@
 package com.gruelbox.transactionoutbox;
 
-
 /** Thrown when de-serialization of an invocation fails. */
 class DeserializationFailedException extends RuntimeException {
   DeserializationFailedException(String message, Throwable cause) {
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/Invocation.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/Invocation.java
@@ -1,10 +1,12 @@
 package com.gruelbox.transactionoutbox;
 
 import com.google.gson.annotations.SerializedName;
+import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.Callable;
 import lombok.Value;
 import lombok.extern.slf4j.Slf4j;
@@ -58,6 +60,11 @@ public class Invocation {
   Map<String, String> mdc;
 
   /**
+   * @return Indicates an Exception during De-Serialization, that is not persisted.
+   */
+  transient Optional<Exception> exceptionDuringDeserialization;
+
+  /**
    * @param className The class name (as provided/expected by an {@link Instantiator}).
    * @param methodName The method name. Combined with {@link #parameterTypes}, uniquely identifies
    *     the method.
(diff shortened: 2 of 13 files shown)


badgerwithagun:
We're also going to need testts for these behaviours!

skastenholz:
Alright, good point. I tried different approaches and came up with the one comitted. In doing so I needed to modify core classes and, in particular, you may have suggestions regarding the Invocation class. Also I added/modified some tests.

Please have a look and provide your feedback. Thanks.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/Invocation.java
```diff
@@ -89,6 +96,19 @@ public Invocation(
...
+  /**
+   * @param exception The exception that occured during de-serialization.
+   */
+  public Invocation(IOException exception) {
+    this.className = "";
+    this.methodName = "";
+    this.parameterTypes = new Class<?>[] {};
+    this.args = new Object[] {};
+    this.mdc = null;
+    this.exceptionDuringDeserialization = Optional.of(exception);
```
badgerwithagun, line 111: Rather than adding a new property here (which will impact on ser/deser of all `Invocation`s, not just this one), why not create a `FailedDeserializingInvocation`, something like this:

```
class FailedDeserializingInvocation extends Invocation {

  private final IOException e;

  FailedDeserializingInvocation(IOException e) {
    this.className = "";
    this.methodName = "";
    this.parameterTypes = new Class<?>[] {};
    this.args = new Object[] {};
    this.mdc = null;
    this.e= e;
  }

  @Override
  void invoke(Object instance, TransactionOutboxListener listener)
      throws NoSuchMethodException, IllegalAccessException, InvocationTargetException {
     throw new WrappedCheckedException(e);
  }

}
```

Should make the code elsewhere cleaner too.

[You will need to replace `@Value`](https://projectlombok.org/features/Value) on `Invocation` to make it nonfinal so this can be done:

```
@ToString
@EqualsAndHashCode
@AllArgsConstructor
@FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
@Getter
```

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DeserializationFailedException.java
```diff
@@ -0,0 +1,8 @@
+package com.gruelbox.transactionoutbox;
+
+/** Thrown when de-serialization of an invocation fails. */
+class DeserializationFailedException extends RuntimeException {
```
badgerwithagun, line 4: I think there's a standard exception somewhere in the core lib for generically wrapping checked exceptions, which is all that's going on here.

transactionoutbox-jackson/src/main/java/com/gruelbox/transactionoutbox/jackson/JacksonInvocationSerializer.java
```diff
@@ -43,24 +43,20 @@ public void serializeInvocation(Invocation invocation, Writer writer) {
...
-      if (checkForOldSerialization(br)) {
-        if (defaultInvocationSerializer == null) {
-          throw new UnsupportedOperationException(
-              "Can't deserialize GSON-format tasks without a "
-                  + DefaultInvocationSerializer.class.getSimpleName()
-                  + ". Supply one when building "
-                  + getClass().getSimpleName());
-        }
-        return defaultInvocationSerializer.deserializeInvocation(br);
+  public Invocation deserializeInvocation(Reader reader) throws IOException {
```
badgerwithagun, line 46: Does this same problem affect `DefaultInvocationSerializer`?

badgerwithagun (changes requested):
This is a really nice approach. I like how neatly it works when you push the exception down into the invocation. Great stuff.

Just a few tweaks on the implementation, but the approach and testing are sound.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/Invocation.java
skastenholz: Yes, makes sense. Done.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DeserializationFailedException.java
skastenholz: Thanks for the hint. Done.

transactionoutbox-jackson/src/main/java/com/gruelbox/transactionoutbox/jackson/JacksonInvocationSerializer.java
skastenholz: According to JavaDoc it could so I applied the same pattern.

badgerwithagun (approved):

badgerwithagun:
Merged! Thanks :)

PR #863, opened 2025-05-19
introduce method of flushing by topic to outbox

victokoh:
(can be extended to flush by topic(s) instead).


badgerwithagun:
Nice idea, thanks @victokoh . Reviewing now

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultPersistor.java
```diff
@@ -339,6 +339,30 @@ public Collection<TransactionOutboxEntry> selectNextInTopics(
...
+  @Override
+  public Collection<TransactionOutboxEntry> selectNextInSelectedTopics(
+      Transaction tx, List<String> topicNames, int batchSize, Instant now) throws Exception {
+    var sql =
+        dialect
+            .getFetchNextInSelectedTopics()
+            .replace("{{table}}", tableName)
+            .replace(
+                "{{topicNames}}",
+                topicNames.stream()
```
badgerwithagun, line 351: Need to use parameters here otherwise this is vulnerable to injection attacks and every call will cause a re-evaluation of the query plan, causing a big performance hit.

```
topicNames.stream().map(it-> "?").collect(Collectors.joining(","))
```

Then

```
topicNames.forEach(tn -> stmt.setString(...));
```
```

File: transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutbox.java [hunk 2]
```diff
@@ -102,6 +102,15 @@ default boolean flush() {
...
   boolean flush(Executor executor);
 
+  /**
+   * Flushes a specific topic
+   *
+   * @param executor
+   * @param topicName
+   * @return
+   */
+  boolean flushTopic(Executor executor, String topicName);
```
badgerwithagun, line 112: Why not allow a collection of topic names?

Perhaps two overloads
```
default boolean flushTopics(Executor executor, String... topicNames) {
  return flushTopics(executor, Arrays.asList(topicNames));
}
boolean flushTopics(Executor executor, Collectio<String> topicNames);
```

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java
```diff
@@ -149,6 +149,36 @@ public boolean flush(Executor executor) {
...
   }
 
+  @Override
+  public boolean flushTopic(Executor executor, String topicName) {
+    if (!initialized.get()) {
+      throw new IllegalStateException("Not initialized");
+    }
+    Instant now = clockProvider.get().instant();
+    List<CompletableFuture<Boolean>> futures = new ArrayList<>();
+    futures.add(
```
badgerwithagun, line 159: Not needed here. `flushTopic` seems very specific and `expireIdempotencyProtection` is unrelated.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutboxImpl.java
```diff
@@ -149,6 +149,36 @@ public boolean flush(Executor executor) {
...
+        CompletableFuture.runAsync(() -> expireIdempotencyProtection(now), executor)
+            .thenApply(it -> false));
+
+    futures.add(
```
badgerwithagun, line 163: With the removal of `expireIdempotencyProtection` you don't need all the `CompletableFuture` stuff.

Just
```
return doFlush(
                  tx ->
                      uncheckedly(
                          () ->
                              persistor.selectNextInSelectedTopics(
                                  tx, topicNames flushBatchSize, now)));
```

transactionoutbox-testing/src/main/java/com/gruelbox/transactionoutbox/testing/AbstractAcceptanceTest.java
```diff
@@ -501,6 +501,63 @@ final void retryBehaviour() throws Exception {
...
+            .submitter(Submitter.withExecutor(singleThreadPool))
+            .attemptFrequency(Duration.ofMillis(500))
+            .listener(processedEntryListener)
+            .blockAfterAttempts(5)
+            .build();
+
+    clearOutbox();
+
+    var selectedTopic = "SELECTED_TOPIC";
+    withRunningFlusher(
```
badgerwithagun, line 524: Rather than using `withRunningFlusher` I think this test would be cleaner to just run a specific `flushTopic` in the calling thread. In practice I don't think you'd ever use `flushTopic` as a looping background thread.

transactionoutbox-testing/src/main/java/com/gruelbox/transactionoutbox/testing/ProcessedEntryListener.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.CountDownLatch;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Collects an ordered list of tx outbox entries that have been 'processed' i.e. succeeded or failed
+ * in processing.
+ */
+@Slf4j
```
badgerwithagun, line 14: Pretty sure this already existed somewhere???!

badgerwithagun (changes requested):
A few small tweaks but looks very sensible.

transactionoutbox-testing/src/main/java/com/gruelbox/transactionoutbox/testing/ProcessedEntryListener.java
victokoh: nope, similar OrderedEntryListener but not quite the same

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutbox.java
victokoh: > (can be extended to flush by topic(s) instead).
I can add it now though

transactionoutbox-testing/src/main/java/com/gruelbox/transactionoutbox/testing/AbstractAcceptanceTest.java
victokoh: It'll still likely occur in a separate thread though. (This particular test is scheduled to fail on the first 3 attempts).

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultPersistor.java
```diff
@@ -339,6 +340,32 @@ public Collection<TransactionOutboxEntry> selectNextInTopics(
     }
   }
 
+  @Override
+  public Collection<TransactionOutboxEntry> selectNextInSelectedTopics(
+      Transaction tx, Collection<String> topicNames, int batchSize, Instant now) throws Exception {
```
badgerwithagun, line 345: With this code you should enforce an _ordered_ collection so I would require a `List`.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/TransactionOutbox.java
```diff
@@ -102,6 +104,20 @@ default boolean flush() {
...
 
+  /**
+   * Flushes a specific topic (or set of topics)
+   *
+   * @param executor to be used for parallelising work (note that the method overall is blocking and
+   *     this is solely ued for fork-join semantics).
+   * @param topicNames the list of specific topics to flush
+   * @return true if any work was flushed
+   */
+  default boolean flushTopic(Executor executor, String... topicNames) {
```
badgerwithagun, line 115: `flushTopics` (plural)

transactionoutbox-testing/src/main/java/com/gruelbox/transactionoutbox/testing/AbstractAcceptanceTest.java
badgerwithagun: Nah I'd call flush three times

badgerwithagun: Out of interest, why are you using a failing instantiator? Is that complicating the test usefully?

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultPersistor.java
victokoh: true

transactionoutbox-testing/src/main/java/com/gruelbox/transactionoutbox/testing/AbstractAcceptanceTest.java
victokoh: a few things are scheduled in the process of this test, a non-ordered task, another task against a separate topic, and lastly - the topic of interest. 

the test confirms that outside of the initial non-ordered task being processed on post-commit, even after some time has passed, only the selected topic ever gets attempted to 'process' by the new flushing method. 

would you still prefer removal of the failing instantiator?

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultPersistor.java
victokoh: it is an array list under the hood btw - but we can make that explicit on the return type of list instead of collection
> var results = new ArrayList<TransactionOutboxEntry>();

transactionoutbox-testing/src/main/java/com/gruelbox/transactionoutbox/testing/AbstractAcceptanceTest.java
badgerwithagun: I'd remove the confusion of failing submissions and background processing, make a completely synchronous test:

1. Submit two ordered tasks in different topics, all of which are guaranteed to run
2. Wait a few seconds and confirm neither have run
3. Flush only one of the topics
4. Confirm only one has run

This test (with the fact that tasks might fail and the use of background processing) is more complex and hard to follow than it needs to be.

There's no point using an unordered task in this test. Whether you `flush`, `flushTopic` or neither of the two, it will definitely run immediately anyway (and there are loads of other tests that confirm this to be the case).

Equally there's no point testing error handling. There's no difference in error handling with your new change compared to existing processing, so there are plenty of other tests checking that and handling potential errors makes your test more complex.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultPersistor.java
victokoh: sure. I think I am a bit confused then (might be missing the obvious!). 

assuming this is about changing Collection<String> topicNames to List<String> topicNames and propagating this all the way up the call hierarchy. 

Q: why do we need to guarantee ordering between the topics? 
this select grabs only the very next item to be processed for each topic. these topics are independent of one another. I would assume then that the order they get processed in exactly doesn't matter.

transactionoutbox-testing/src/main/java/com/gruelbox/transactionoutbox/testing/AbstractAcceptanceTest.java
victokoh: sure, thank you. I have amended the test to reflect this discussion / it is now fully synchronous.

transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultPersistor.java
victokoh: I see, so this is about desiring predictability in the data that we pass in.
I will make the change, however, this data is being used to populate an IN clause under the hood, within a sql query that does not declare an absolute ordering on the results it returns. We make no promises then on a guaranteed result returned from this method.

badgerwithagun: Fair point actually, since `?`s don't need to be in any particular order! Sorry I'm being thick.

badgerwithagun (approved):

PR #864, opened 2025-05-22
Retry options (WIP)

victokoh:


goulashify:
+1, this would be useful for us and would make this lib even more flexible! Our use case is having shorter retries for immediate actions (like seconds when process payments through 3rd parties while the user is waiting) while keeping long-ones that can retry for long in extreme cases (like days for a reconciliation process where a 3rd party is down) and try to self-heal.

PR #866, opened 2025-05-23
Ordered batch support

indriesergiu:
This PR aims to increase the throughput of the transactional outbox by implementing batch processing for ordered items.
This significantly improves performance while maintaining ordering guarantees within topics.

## Details

- **Ordered Batch Processing**: Process items in batches while maintaining order within topics
- **Parallel Execution**: Run different topic batches in parallel
- **Efficient Locking**: Batch locking of entries to improve performance
- **Error Handling**: Comprehensive error handling for batch failures
- **Listener Notifications**: Properly notify listeners of success/failure events
- **Testing**: Dedicated tests to verify functionality

## Batch Error Handling

The implementation uses an all-or-nothing approach for error handling in batch processing. When any entry in a batch fails during execution, the entire batch is marked for retry, maintaining strict ordering guarantees within topics. This approach ensures transactional consistency at the cost of potentially reduced throughput when problematic entries exist. Upon failure, the system increments attempt counts for all entries in the batch, triggers appropriate failure listeners, and rolls back the transaction. This guarantees that entries within a topic are always processed in sequence, even after failures. The batch will be retried according to the configured retry schedule until successful completion or until entries reach their maximum attempt threshold and become blocked. This simpler approach prioritizes correctness and ordering over partial processing, making the system more predictable when handling failures.

## Important Notes
1. I've temporarily disabled 2 tests for (MS SQL and MySQL) since they're not a priority for us, but this can be solved moving forward
2. See details regarding the batch error handling above. The implemented behavior is a simple solution that matches our needs, but we can discuss changes.
3. I avoided adding a batch method to the `Submitter` interface and worked around that by using the same `Executor`. I assume we will have to change this in some way.

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,0 +1,1 @@
+-Dmaven.repo.local=/workspace/.m2 
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,1 +1,0 @@
--Dmaven.repo.local=/workspace/.m2 
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <skip.format>false</skip.format>
     <!-- Dependency versions -->
     <logback.version>1.5.18</logback.version>
-    <revision>1.3.99999-SNAPSHOT</revision>
+    <revision>5.5.99999-SNAPSHOT</revision>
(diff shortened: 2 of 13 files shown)


indriesergiu:
Cool, thanks!
Waiting to hear from you :)

PR #870, opened 2025-05-26
gradle project settings

oriy:
dependabot would be able to upgrade both maven and gradle simultaneously


oriy:
> Not intending to start a Gradle vs Maven conversation; they are both valid options. However, what's the point of using both in the same project? Does it help in some way? We use transaction-outbox in both Maven and Gradle projects seamlessly. Dependabot works for both.

No point keeping both
Gradle currently misses the executing per DB
Suggests better caching and faster execution

PR #930, opened 2025-10-19
Add trace context to invocation

mfilatov:
I would like to propose the following changes to support the storage and recovery of trace information.

It is based on an earlier discussion #862 that has not yet closed.


badgerwithagun:
I've gone for a more open solution that is less obviously tied to OTEL. See #931 

This makes OTEL propagation easy in your application while not being opinionated about how you choose to do it, and also allows for other session information which isn't OTEL related; the sort of thing that the MDC might have previously been abused for.

badgerwithagun:
@mfilatov that's in `master` now. Does it work for you?

PR #945, opened 2025-10-22
Some changes to support session serialization and deserialization

mfilatov:
Some changes to support session serialization and deserialization in transactionoutbox-jackson module.
Also the constructor of the Invocation class has been replaced with a new one.


badgerwithagun:
I'm working on this at the moment @mfilatov , no need to double-up effort

PR #1060, opened 2026-06-29
- Fix proxy interception for inherited methods

pfivanov:
## Problem
When proxying concrete classes, inherited methods were not intercepted by
ByteBuddy. Calls executed directly on Objenesis-created instances **without constructor invocation**, leaving superclass fields as `null`.
## Solution
- ProxyFactory: Match methods from entire class hierarchy, not just the target class
- Invocation: Walk superclass chain to find inherited methods
## Impact
Fixes NPE failures in classes extending abstract base classes.


transactionoutbox-spring/src/test/java/com/gruelbox/transactionoutbox/spring/it/SpringTransactionManagerIT.java
```diff
@@ -43,4 +46,23 @@ public void shouldThrowAlreadyScheduledException() {
           return null;
         });
   }
+
```
badgerwithagun: Rather than testing this in the spring module, I'd prefer a unit test in the core module. You can add it to `TestProxyGeneration`.

badgerwithagun (changes requested):
Thanks @pfivanov!

The change looks good. Just a small change to the tests needed.

transactionoutbox-spring/src/test/java/com/gruelbox/transactionoutbox/spring/it/SpringTransactionManagerIT.java
pfivanov: The test has been moved to `TestProxyGeneration` in the core module.

PR #1066, opened 2026-07-13
Don't popTransaction unless it was pushed

### LINKED ISSUE #1065: SimpleTransactionManager.withTransaction calls popTransaction() even when pushTransaction was never reached
### Summary

In `SimpleTransactionManager.withTransaction` (`transactionoutbox-core/.../SimpleTransactionManager.java:56`), `popTransaction()` is called unconditionally in the `finally` block, but the matching `pushTransaction(...)` happens as the second resource in the try-with-resources. If the first resource — `connectionProvider.obtainConnection()` — throws, the transaction is never pushed, yet `popTransaction()` still runs.
```java
try (Connection connection = connectionProvider.obtainConnection();   // (1) if this throws...
     SimpleTransaction transaction = pushTransaction(new SimpleTransaction(connection, null))) {  // (2) ...this never runs
    ...
} catch (SQLException e) {
    throw new RuntimeException(e);
} finally {
    popTransaction();   // ...but this always runs
}
```
### Impact
`popTransaction()` calls `transactions.get().pop()` on the thread-local deque (`AbstractThreadLocalTransactionManager:52`). Because nothing was pushed, one of two things happens:
1. No enclosing transaction: the deque is empty, so `pop()` throws `NoSuchElementException` from the `finally` block. This masks the original exception from `obtainConnection()` (e.g. the real "cannot get connection" / pool-exhausted `SQLException`), so callers see a confusing `NoSuchElementException` instead of the root cause.
2. Nested/enclosing transaction present on the same thread: `pop()` removes the parent's transaction from the stack, corrupting the thread-local state of the enclosing transaction.

### Steps to reproduce

Configure a `SimpleTransactionManager` with a `ConnectionProvider` that throws on `obtainConnection()` (e.g. simulate a connection-pool timeout), then call any `inTransaction*` method. Observe that a `NoSuchElementException` is thrown rather than the underlying connection failure.

### Suggested fix

Only pop if a push actually succeeded — e.g. move pushTransaction out of the try-with-resources so push/pop are balanced, or guard the finally.

MrLIVB:
Fix #1065 

Reorganizes code inside of `SimpleTransactionManagerTest.withTransaction` so `popTransaction` isn't called in `finally` block of connection acquiring.

### DIFF
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/SimpleTransactionManager.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/SimpleTransactionManager.java
@@ -56,22 +56,22 @@ final class SimpleTransactionManager
   private <T, E extends Exception> T withTransaction(ThrowingTransactionalSupplier<T, E> work)
       throws E {
     try (Connection connection = connectionProvider.obtainConnection();
-        SimpleTransaction transaction = pushTransaction(new SimpleTransaction(connection, null))) {
+        SimpleTransaction transaction = new SimpleTransaction(connection, null)) {
       log.debug("Got connection {}", connection);
-      boolean autoCommit = transaction.connection().getAutoCommit();
+      boolean autoCommit = connection.getAutoCommit();
       if (autoCommit) {
         log.debug("Setting auto-commit false");
-        Utils.uncheck(() -> transaction.connection().setAutoCommit(false));
+        Utils.uncheck(() -> connection.setAutoCommit(false));
       }
+      pushTransaction(transaction);
       try {
         return work.doWork(transaction);
       } finally {
+        popTransaction();
         connection.setAutoCommit(autoCommit);
       }
     } catch (SQLException e) {
       throw new RuntimeException(e);
-    } finally {
-      popTransaction();
     }
   }
 }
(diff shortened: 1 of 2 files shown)


transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/SimpleTransactionManager.java
```diff
@@ -56,22 +56,22 @@ private <T, E extends Exception> T processAndCommitOrRollback(
...
       if (autoCommit) {
         log.debug("Setting auto-commit false");
-        Utils.uncheck(() -> transaction.connection().setAutoCommit(false));
+        Utils.uncheck(() -> connection.setAutoCommit(false));
       }
+      pushTransaction(transaction);
       try {
         return work.doWork(transaction);
       } finally {
+        popTransaction();
```
badgerwithagun, line 70: Now we need to be similarly careful about `connection.setAutoCommit(autoCommit);`. If `pushTransaction` or `popTransaction` throw, the connection will be left with a corrupted autocommit setting. Another nested `try...finally` is needed.

badgerwithagun (changes requested):
Thanks @MrLIVB! Good spot. There's a small tweak needed then this is good to merge.

PR #1068, opened 2026-07-13
Fix race condition on PostgreSQL when assigning first sequence number to 'TransactionOutboxEntry'

### LINKED ISSUE #1067: Race condition on PostgreSQL when assigning first sequence number to 'TransactionOutboxEntry'
Hello, 

The way `DefaultPersistor#setNextSequence` handles race condition to assign a first sequence number does not work with PostgreSQL. The same connection is re-used for a recursive call when an `indexViolation` has been detected. This is not allowed by PostgreSQL and will lead to the following error :  
```
ERROR:  current transaction is aborted, commands ignored until end of transaction block
```
As a result, scheduling two executions at the same time (on the same topic) could fail for one execution.

gregoire-rouillon:
Fix #1067

### DIFF
--- a/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultDialect.java
+++ b/transactionoutbox-core/src/main/java/com/gruelbox/transactionoutbox/DefaultDialect.java
@@ -31,6 +31,7 @@ class DefaultDialect implements Dialect {
   @Getter private final String fetchNextInSelectedTopics;
   @Getter private final String fetchCurrentVersion;
   @Getter private final String fetchNextSequence;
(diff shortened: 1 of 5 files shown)


badgerwithagun (changes requested):
Hi @gregoire-rouillon, welcome and thank you for your PR.

This one looks interesting. What we really need is a demonstration test which replicates the issue. `TestPostgres16` (or 15, or 14...) may be the best place for it.

There are quite a few examples in that stack of mimicking race conditions using listeners and concurrency primitives like `CountdownLatch`. Let me know if you need a suggestion.

gregoire-rouillon:
Hi @badgerwithagun, here is the requested change. Adding a test has highlighted another issue, but this time with MySQL.

PR #1070, opened 2026-07-18
MySQL: restore blocked column to BOOLEAN so IX_TXNO_OUTBOX_1 is usable

andreib-goji:
**Problem:** migration 6 renamed blacklisted to blocked but also changed the MySQL column type from `BOOLEAN` to `VARCHAR(250)`. Boolean comparisons against the `VARCHAR` need an implicit cast, so the retention cleanup `DELETE` can't use blocked as an index range condition on `IX_TXNO_OUTBOX_1` and scans the whole `processed = true` region. Under `REPEATABLE READ` it `next-key-locks` every retained row it scans (even when it deletes nothin) and concurrent outbox writers hit lock wait timeouts.

**Fix:** migration 14 (MySQL dialects only) converts the column back to `BOOLEAN`. Existing "0"/"1" values convert in place. Queries keep boolean literals.

Tests:
- `TestMySqlDialect` pins migration 14 SQL and boolean comparisons.
- `migration14ConvertsExistingBlockedValues` rebuilds a populated v13 schema by replaying real migrations, then migrates and verifies type and semantics (MySQL 5 + 8).
- `retentionCleanupDoesNotLockRetainedRecords` proves an idle cleanup transaction no longer blocks writes to retained rows. Passes on H2, MySQL 5/8, PostgreSQL 16, Oracle 21, SQL Server 2019.

### DIFF
--- a/transactionoutbox-acceptance/src/test/java/com/gruelbox/transactionoutbox/acceptance/persistor/AbstractMySqlPersistorTest.java
+++ b/transactionoutbox-acceptance/src/test/java/com/gruelbox/transactionoutbox/acceptance/persistor/AbstractMySqlPersistorTest.java
@@ -1,0 +1,114 @@
+package com.gruelbox.transactionoutbox.acceptance.persistor;
+
+import static java.time.temporal.ChronoUnit.MILLIS;
+import static java.util.stream.Collectors.toSet;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import com.gruelbox.transactionoutbox.Invocation;
+import com.gruelbox.transactionoutbox.Persistor;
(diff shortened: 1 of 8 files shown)


badgerwithagun (changes requested):
Hi @andreib-goji and thank you for your PR.

Looking at migration 6, I actually think this was an error in the PR that introduced the schema change. `blocked` shouldn't be a varchar. The correct change here will be to add a migration to return the column to the correct type (correctly migrating existing values) rather than work around it.

We also need an acceptance test (in https://github.com/gruelbox/transaction-outbox/blob/master/transactionoutbox-testing/src/main/java/com/gruelbox/transactionoutbox/testing/AbstractAcceptanceTest.java) to ensure the original issue is resolved on all dialects. You need to replicate the failed blocking symptoms there before applying your fix.

